PR #10, opened 2019-06-13
Transfers: send event when recipient is called

sduthil:
reason: client app needs the recipient_call as soon as possible


pc-m:
merged thanks

PR #14, opened 2019-06-26
applications: add /answer on calls

sduthil:


wazo_calld/plugins/applications/events.py
```diff
@@ -111,6 +111,11 @@ class CallUpdated(_BaseCallItemEvent):
     routing_key = 'applications.{}.calls.{}.updated'
 
 
+class CallAnswered(_BaseCallItemEvent):
```
sboily: Why there is no doc for this event with asyncAPI?

sduthil:
Totally agree about AsyncAPI docs. Did not want to block the PR just for that.

fblackburn1:
Some integration tests fail:
ERROR: test_entering_stasis_without_a_node (suite.test_applications.TestStasisTriggers) 
ERROR: test_entering_stasis_with_a_node (suite.test_applications.TestStasisTriggers)
ERROR: test_post_node_call (suite.test_applications.TestApplication)
ERROR: test_post_call (suite.test_applications.TestApplication)

PR #17, opened 2019-06-28
applications: call user into node

sduthil:


fblackburn1:
Should we validate that user_uuid exists before returning 201 or it doesn't worth the performance cost

PR #19, opened 2019-07-01
Add push mobile plugin

sileht:


wazo_calld/plugins/mobile/bus_consume.py
```diff
@@ -0,0 +1,36 @@
...
+class PushNotificationBusEventHandler(object):
+
+    def __init__(self, bus_publisher):
+        self.bus_publisher = bus_publisher
+
+    def subscribe(self, bus_consumer):
+        bus_consumer.on_ami_event('UserEvent', self._user_event)
+
+    def _user_event(self, event):
+        if event['UserEvent'] == 'Pushmobile':
```
pc-m: An early return would be nice here

sileht: fixed

PR #20, opened 2019-07-01
switchboard: allow to select line to answer

sileht:


sileht:
> Wouldn't it be better to have the `line_id` in the query string instead of in the body? The rationale behind this is that a PUT on a resource with a body usually modifies something. In this case it only selects the line. What do you think?

I really don't have any strong opinion about this one as this endpoint is not about modify resources, I'm ok with both solution.

sileht:
> Wouldn't it be better to have the `line_id` in the query string instead of in the body? The rationale behind this is that a PUT on a resource with a body usually modifies something. In this case it only selects the line. What do you think?

Done

integration_tests/suite/test_switchboards.py
```diff
@@ -431,6 +436,65 @@ def answered_event_received():
...
+
+        # Verify second line is used
+        channels = self.ari.channels.list()
+        assert_that(len(channels), equal_to(2))
+        channel_answer = [c for c in channels if c.id != queued_call_id][0]
+        assert_that(
+            channel_answer.json.get('name'),
+            starts_with('Test/switchboard-second-line')
+        )
+
```
bloom1: Would it be useful to create a method or a hamcrest matcher to avoid duplication?

bloom1: I forgot: an entry describing the addition of the query string in the CHANGELOG.md is missing.

sileht: > Would it be useful to create a method or a hamcrest matcher to avoid duplication?

It's a bit specific to switchboard answering, not sure it worses it.

sileht: > I forgot: an entry describing the addition of the query string in the CHANGELOG.md is missing.

done

PR #23, opened 2019-07-03
application: add /progress start and stop

sduthil:


wazo_calld/plugins/applications/api.yml
```diff
@@ -220,6 +220,51 @@ paths:
         '503':
           $ref: '#/responses/AnotherServiceUnavailable'
 
+  /applications/{application_uuid}/calls/{call_id}/progress/start:
+    put:
+      summary: Play the progress ringing tone
+      description: >
```
fblackburn1: Why using `>` instead `|` as we use everywhere?

wazo_calld/plugins/applications/api.yml
```diff
@@ -220,6 +220,51 @@ paths:
...
+
+  /applications/{application_uuid}/calls/{call_id}/progress/stop:
+    put:
+      summary: Stop playing the progress ringing tone.
+      description: >
+        **Required ACL:** `calld.applications.{application_uuid}.calls.{call_id}.progress.stop.update`
+
+        The progress ringing tone is the sound heard by a caller while the other party is ringing.
+
+        Limitation: this only works when the ringing tone is played by Wazo through audio media (e.g. when using `/progress` after `/answer`). If the phone is emitting the ringing tone directly to the user, then it cannot be stopped, and the response code will still be 204.
```
fblackburn1: objection to wrap line?

wazo_calld/plugins/applications/api.yml
```diff
@@ -620,6 +665,16 @@ definitions:
         type: string
       status:
         type: string
+        description: >
```
fblackburn1: using `|` no need to have empty line between points

PR #27, opened 2019-07-11
token_renewer: use new callback signature

sileht:
token_renewer now passes the whole token object to callbacks


sileht:
Closed in favor of https://github.com/wazo-pbx/xivo-lib-python/pull/21

PR #29, opened 2019-07-17
applications: snoop list: dont fail when snoop is being created

sduthil:


wazo_calld/plugins/applications/models.py
```diff
@@ -177,10 +183,12 @@ def new_snoop_channel(self, ari, whisper_mode):
         snoop_channel.setChannelVar(
             variable=self._whisper_mode_chan_var,
             value=whisper_mode,
+            bypassStasis=True,
```
fblackburn1: Why? I though all channels from application are in stasis

sduthil: Me too, but on Jenkins, we get status code 409, meaning "Channel not in Stasis". I guess there is a slight delay between the REST API return and the new channel being put in Stasis.

sduthil: Ideally we should only execute the `setChannelVar` on `StasisStart` event from the `snoop_channel`, but it would get rather hard to follow then.

PR #43, opened 2019-11-05
switchboards: implement cache for getting switchboards in wazo-confd

sduthil:
reason: all requests to /switchboards are slowed down by those requests
to wazo-confd, where the response does not change frequently


sduthil:
Superseded by #44

PR #32, opened 2019-08-12
add zuul jobs

fredericlepied:
Depends-On: https://github.com/wazo-platform/sf-jobs/pull/15

PR #54, opened 2019-12-26
snoop: wait until the snoop channel gets into status on snoop create

pc-m:
on slow computers it is possible that the channel gets added to the bridge
before getting into stasis. This results in a 422 error from ARI and undefined
and a 500 error from calld.

I also removed the bypassStasis arguments since they were used to work around
this issue.


pc-m:
The Zuul PR should be fixed with this branch https://github.com/wazo-platform/wazo-calld/pull/32

pc-m:
> The Zuul PR should be fixed with this branch #32

Yay it worked!

PR #56, opened 2019-12-27
Use real calld client

fblackburn1:
Depends-on: ~~https://github.com/wazo-platform/wazo-calld/pull/50~~
Depends-on: ~~https://github.com/wazo-platform/wazo-calld-client/pull/17~~
Depends-on: https://github.com/wazo-platform/wazo-calld-client/pull/19


fblackburn1:
recheck

PR #60, opened 2020-01-07
dial_mobile: fix logger which misses an argument

pc-m:


pc-m:
Depends-On: https://github.com/wazo-platform/wazo-calld/pull/61

bloom1:
recheck

PR #63, opened 2020-01-13
WAZO-1476: do not git pull in integration_tests Makefile

fredericlepied:


fblackburn1:
@fredericlepied Can we set the `INTEGRATION_TEST_TIMEOUT` by default for all zuul job (something like 60)

PR #64, opened 2020-01-20
Make ssl optional

safchain:


wazo_calld/service_discovery.py
```diff
@@ -3,10 +3,17 @@
 
 import requests
 
+from wazo_calld.http_server import VERSION
+
 
 # this function is not executed from the main thread
-def self_check(port):
-    url = 'https://localhost:{}/1.0/status'.format(port)
+def self_check(config):
```
pc-m: This function is still called with the port only in controller.py

safchain: oups, thx for pointing this

PR #65, opened 2020-01-25
Wazo 1501 mute unmute api for calls

fblackburn1:
Depends-on: https://github.com/wazo-platform/wazo-calld-client/pull/20


fblackburn1:
recheck

PR #69, opened 2020-02-06
calls: add call_answered event

sduthil:
Why:

* the call_updated event does not allow to detect exactly when a call
has been answered because it may be sent multiple times after the call
is answered.

### DIFF
--- a/integration_tests/suite/helpers/bus.py
+++ b/integration_tests/suite/helpers/bus.py
@@ -1,4 +1,4 @@
-# Copyright 2015-2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2015-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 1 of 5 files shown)


sboily:
recheck

bloom1:
recheck

bloom1:
recheck

integration_tests/suite/helpers/bus.py
```diff
@@ -60,11 +60,12 @@ def send_ami_newchannel_event(self, channel_id, channel=None):
             }
         }, 'ami.Newchannel')
 
-    def send_ami_newstate_event(self, channel_id):
+    def send_ami_newstate_event(self, channel_id, state='Up'):
         self.send_event({
             'data': {
                 'Event': 'Newstate',
                 'Uniqueid': channel_id,
+                'State': 'Up',
```
pc-m: This should be `state` not `Up`

sduthil:
recheck

PR #74, opened 2020-02-19
WAZO-1337 : add auto-answer headers to relocates and calls

bloom1:


wazo_calld/plugins/calls/api.yml
```diff
@@ -315,6 +315,9 @@ definitions:
       all_lines:
         type: boolean
         description: Use all of the lines of the user to make the call (ignored when `line_id` is specified).
+      auto_answer:
```
sduthil: Thins is ambiguous: when `POST /calls` is used, there are two legs who are ringing: first the caller, then the callee. This setting only applies to the caller leg. The key name should reflect this behavior.

wazo_calld/plugins/relocates/state.py
```diff
@@ -88,13 +88,14 @@ class RelocateStateReady(RelocateState):
 
     def initiate(self, relocate, destination):
         endpoint = destination.ari_endpoint()
+        variables = relocate.recipient_variables or {}
```
sduthil: Can you initialize `recipient_variables` to `{}` instead?

PR #80, opened 2020-04-03
remove ctid-ng compatibility reverse proxy

fblackburn1:


fblackburn1:
> _Warning_: We are still use the ctid-ng URL in our js-sdk

not anymore https://wazo-dev.atlassian.net/browse/WDA-1196

PR #81, opened 2020-04-09
adding "none" to accepted values

sutidor:
According to https://wazo-platform.org/documentation/console/application
'none' is a valid value


wazo_calld/plugins/applications/schemas.py
```diff
@@ -77,7 +77,7 @@ class ApplicationNodeSchema(BaseSchema):
 
 
 class ApplicationSnoopPutSchema(BaseSchema):
-    whisper_mode = fields.String(validate=OneOf(['in', 'out', 'both']), missing=None)
+    whisper_mode = fields.String(validate=OneOf(['in', 'out', 'both', 'none']), missing=None)
```
pc-m, line 80: Could you also change the `missing` parameter from `None` to `'none'`. This would avoid having 2 possible values doing the same thing.

fblackburn1: If doing this, we also need to remove the `load_whisper_mode` and `dump_whisper_mode` method to avoid confusion
IMO, the API has been done to accept 3 different string values (in/out/both) and the None value. But it seems to have a little bug (a missing `allow_none=True`) to accept the `whisper_mode=None`.
And we also need to fix the documentation
Because for now to set the `whisper_mode` to none you need to remove the key (which is annoying)

sutidor: I hope I got it right now according to your comments. Otherwise I dont mind if you reject my pull request and do the changes for yourselves.

PR #89, opened 2020-05-06
WAZO-1370: fix 500 error when user has no voicemail

bloom1:
Depends-On: https://github.com/wazo-platform/wazo-calld/pull/90

### DIFF
--- a/wazo_calld/plugin_helpers/confd.py
+++ b/wazo_calld/plugin_helpers/confd.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2016-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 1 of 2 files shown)


wazo_calld/plugin_helpers/confd.py
```diff
@@ -132,7 +132,10 @@ def from_id(cls, conference_id, confd_client):
 
 def get_user_voicemail(user_uuid, confd_client):
     try:
-        return confd_client.users.get(user_uuid)['voicemail']
+        voicemail = confd_client.users.get(user_uuid)['voicemail']
+        if not voicemail:
+            raise NoSuchUserVoicemail(user_uuid)
+        return voicemail
     except IndexError:
```
pc-m, line 139: Is the `IndexError` still useful here? I feel that it was meant to catch a user with no `voicemail` field.

bloom1: Probably not. I will remove it. Thanks!

bloom1:
recheck

PR #90, opened 2020-05-08
WAZO-1762 remove foreground option from amid

fblackburn1:
reason: we also have removed the -f option from amid, which is highlight
this issue


pc-m:
recheck

PR #91, opened 2020-05-23
reduce docker image size

fblackburn1:


Dockerfile
```diff
@@ -1,27 +1,36 @@
...
+COPY ./bin/wazo-pdf2fax /usr/bin/wazo-pdf2fax
 
 RUN true \
-    && apt-get -qq update \
-    && apt-get -qq -y install libpq-dev libyaml-dev ghostscript \
-    && pip install -r requirements.txt \
-    && python setup.py install \
-    && cp -av etc/wazo-calld /etc \
-    && install -m 755 -o root -g www-data bin/wazo-pdf2fax /usr/bin/wazo-pdf2fax \
+    && apt-get -q update \
```
bloom1, line 23: Is the `DEBIAN_FRONTEND` environment variable necessary in the first image or this one? It is defined in the first image but `apt-get` is used in the second image.

fblackburn1: nice catch, well since it doesn't seem to change anything, I removed it

PR #92, opened 2020-05-29
Pull request for WAZO-1795-calld-without-https

fblackburn1:
debian: remove unused postinst
use http by default and deprecate https
remove unused verify for consul


fblackburn1:
regate

PR #103, opened 2020-07-17
conference: ignore joins for unknown conferences

sduthil:
Why:

* Pagings create confbridge conferences that are not found in
wazo-confd.
* We want to avoid tracebacks when that happens.

### DIFF
--- a/wazo_calld/plugin_helpers/confd.py
+++ b/wazo_calld/plugin_helpers/confd.py
@@ -122,7 +122,7 @@ class Conference:
         try:
             conference = confd_client.conferences.get(conference_id)
         except HTTPError as e:
-            if e.response and e.response.status_code == 404:
+            if e.response is not None and e.response.status_code == 404:
                 raise NoSuchConferenceID(conference_id)
             raise
         return cls(conference['tenant_uuid'],
--- a/wazo_calld/plugins/conferences/bus_consume.py
+++ b/wazo_calld/plugins/conferences/bus_consume.py
@@ -4,6 +4,7 @@
 import logging
 
 from wazo_calld.plugin_helpers.confd import Conference
+from wazo_calld.plugin_helpers.exceptions import NoSuchConferenceID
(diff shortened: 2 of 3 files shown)


wazo_calld/plugin_helpers/confd.py
```diff
@@ -122,7 +122,7 @@ def from_id(cls, conference_id, confd_client):
         try:
             conference = confd_client.conferences.get(conference_id)
         except HTTPError as e:
-            if e.response and e.response.status_code == 404:
+            if e.response is not None and e.response.status_code == 404:
```
fblackburn1, line 125: Out of curiosity, what was the issue with `e.response` only?

wazo_calld/plugins/conferences/bus_consume.py
```diff
@@ -28,6 +29,12 @@ def subscribe(self, bus_consumer):
 
     def _notify_participant_joined(self, event):
         conference_id = int(event['Conference'])
+        try:
+            conference = Conference.from_id(conference_id, self._confd)
```
bloom1, line 33: Would it be useful to add tests for this?

fblackburn1: :-1: for integration tests, we should not add test if there is no functional impact and be careful with what we tests. Testing too much useless things can be worst than testing nothing
But no objection for unittests

sduthil: Agreed for functional test. I don't like testing for logs, and testing for a non-event is too slow because you have to wait for the timeout. I asked myself the same question for unit-tests, so I guess it's worth it :) I'll add the test.

fblackburn1: For this fix, I prefer to no have test than blocking the PR while 7 months :shrug:

fblackburn1 (approved):

PR #109, opened 2020-08-06
[WAZO-1904] Define if the media call is video

sboily:


fblackburn1:
Fixed by https://github.com/wazo-platform/wazo-calld/pull/130

PR #112, opened 2020-08-13
line/trunk status: get the username from the auth_section_options

pc-m:
username is not a top level field of an endpoint SIP anymore.

### DIFF
--- a/wazo_calld/plugins/endpoints/services.py
+++ b/wazo_calld/plugins/endpoints/services.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2019-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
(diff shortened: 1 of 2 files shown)


wazo_calld/plugins/endpoints/services.py
```diff
@@ -248,7 +248,12 @@ def _update_trunk_cache(self, trunks):
             if trunk.get('endpoint_sip'):
                 techno = 'sip'
                 name = trunk['endpoint_sip']['name']
-                username = trunk['endpoint_sip']['username']
+                for key, value in trunk['endpoint_sip'].get('auth_section_options', []):
```
fblackburn1, line 251: auth_section_options is part of the spec
```suggestion
                for key, value in trunk['endpoint_sip']['auth_section_options']:
```

fblackburn1:
alternative for fun
```python
options = trunk['endpoint_sip']['auth_section_options'] 
try:
    username = next(value for key, value in options if key == 'username')
except StopIteration:
    username = None

PR #115, opened 2020-08-17
Pull request for WAZO-1862-adhoc-conferences

sduthil:
## adhoc conferences: rename dialplan entry point


sduthil:
recheck

sduthil:
recheck

wazo_calld/plugins/adhoc_conferences/stasis.py
```diff
@@ -0,0 +1,53 @@
...
+            logger.debug('ignoring StasisStart event: channel %s, app %s, args %s',
+                         event['channel']['name'],
+                         event['application'],
+                         event['args'])
+            return
+
+        if sub_app != 'adhoc_conference':
+            return
+
+        with self.adhoc_conference_creation_lock:
```
pc-m, line 42: I don't think the lock is necessary here if you switch the createWithId with the get.
First try to create it, `get` it if it already exists.
If it's gone by the time you `get` it panic and hangup.

wazo_calld/plugins/adhoc_conferences/stasis.py
```diff
@@ -0,0 +1,53 @@
...
+                bridge = self.ari.bridges.get(bridgeId=adhoc_conference_id)
+            except ARINotFound:
+                logger.debug('adhoc conference %s: creating bridge', adhoc_conference_id)
+                bridge = self.ari.bridges.createWithId(
+                    type='mixing',
+                    bridgeId=adhoc_conference_id,
+                )
+        channel_id = event['channel']['id']
+        logger.debug('adhoc conference %s: bridging participant %s', adhoc_conference_id, channel_id)
+        bridge.addChannel(channel=channel_id)
```
pc-m, line 53: Do we need the thing here to avoid messing up the connected line ids of everyone in the bridge? I do not have the answer to that question, the `addChannel` made me think of that problem.

PR #118, opened 2020-08-20
Pull request for WAZO-1917-adhoc-conference-add-remove-participant

sduthil:
Depends-On: https://github.com/wazo-platform/wazo-calld-client/pull/31
Depends-On: wazo-platform/xivo-bus#25
Depends-On: https://github.com/wazo-platform/wazo-calld/pull/117


sduthil:
recheck

wazo_calld/plugins/adhoc_conferences/notifier.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def participant_joined(self, adhoc_conference_id, other_participant_user_uuids, participant_call):
+        for other_participant_user_uuid in other_participant_user_uuids:
+            participant_call_body = call_schema.dump(participant_call)
+            self._bus_producer.publish(
+                AdhocConferenceParticipantJoinedUserEvent(
+                    adhoc_conference_id,
+                    other_participant_user_uuid,
+                    participant_call_body,
+                )
```
bloom1, line 40: I'm not sure if I understand correctly, but why is the event published without the `user_uuid` header here and for `participant_left`?

sduthil: I wanted to send only one event with the user headers of all participants in one shot, but I forgot to do it. I made a simpler version where each event sent in a routing key also bears one user header.

PR #121, opened 2020-08-21
WAZO-1926: add call answer

bloom1:
Depends-On: https://github.com/wazo-platform/wazo-calld-client/pull/33

### DIFF
--- a/wazo_calld/plugins/calls/api.yml
+++ b/wazo_calld/plugins/calls/api.yml
@@ -411,7 +411,7 @@ paths:
       - users
       responses:
         '204':
-          description: The call is now unheld
+          description: The call is now answered
         '404':
           description: No such call
           schema:
--- a/wazo_calld/plugins/calls/api.yml
+++ b/wazo_calld/plugins/calls/api.yml
@@ -131,7 +131,6 @@ paths:
       description: '**Required ACL:** `calld.calls.{call_id}.hold.start.update`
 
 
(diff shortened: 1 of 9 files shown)


wazo_calld/plugins/calls/api.yml
```diff
@@ -377,6 +397,27 @@ paths:
...
+
+        Only calls owned by the authenticated user may be answered.'
+      parameters:
+      - $ref: '#/parameters/CallID'
+      tags:
+      - calls
+      - users
+      responses:
+        '204':
+          description: The call is now unheld
```
pc-m, line 414: `s/unheld/answered/`

wazo_calld/plugins/calls/services.py
```diff
@@ -335,6 +335,20 @@ def unhold_user(self, call_id, user_uuid):
...
 
+    def answer(self, call_id):
+        try:
+            channel = self._ari.channels.get(channelId=call_id)
+        except ARINotFound:
+            raise NoSuchCall(call_id)
+
+        protocol_interface = protocol_interface_from_channel(channel.json['name'])
+
+        self._phoned_client.answer_endpoint(protocol_interface.interface)
```
pc-m, line 346: phoned could could return `400` no plugin, `404` no such endpoint or `500`. We should do something here to avoid returning `500` 

I think there are some case of exception that wraps other service errors to return a 503 with the original error.

PR #125, opened 2020-08-31
Pull request for WAZO-1929-adhoc-conference-refactor

sduthil:
## adhoc conference: rename definitions

Why:

* Homogenize naming

## adhoc conference tests: change participant_count

Why:

* More symmetric with API definition

## adhoc conference tests: make preconditions independent of calld client

Why:

* Preconditions should be usable in any order
* If a test modifies self.calld_client, it should not fail the
precondition

## adhoc conference: remove duplicate method find_call_peer


fblackburn1 (dismissed):
Thanks
`make_user_calld` :heart:

fblackburn1 (approved):

PR #128, opened 2020-09-08
dial_mobile: automatically hang up channel left alone

sduthil:
Why:

* When in a 1 on 1 conversation, if one party hangs up, the other party
should be hungup too.
* Automatic hangup has been restricted to the callcontrol Stasis
application in 2ed02320913c721eab126ab8f7272b523469da73


pc-m (approved):

sduthil:
I merged manually to avoid waiting another 1h15 from zuul...

PR #138, opened 2020-12-15
test: amid config need to have auth username/password

fblackburn1:
to generate its refreshtoken to know its master tenant for REST API

Depends-on: https://github.com/wazo-platform/wazo-amid/pull/31


pc-m (approved):

pc-m:
recheck

PR #141, opened 2021-01-14
WAZO-2171: call record stop

bloom1:
Depends-On: https://github.com/wazo-platform/wazo-calld-client/pull/35

You should also check the wazo-auth-keys PR: https://github.com/wazo-platform/wazo-auth-keys/pull/33


bloom1:
recheck

PR #143, opened 2021-01-18
calls: send call_updated when recording state changes

sduthil:


fblackburn1:
I think we should remove:
* https://github.com/wazo-platform/wazo-calld/blob/master/wazo_calld/plugins/calls/services.py#L377-L380
* https://github.com/wazo-platform/wazo-calld/blob/master/wazo_calld/plugins/calls/services.py#L394-L397

fblackburn1 (dismissed):

fblackburn1 (dismissed):

PR #145, opened 2021-01-19
Raise error if recordfile not set

fblackburn1:


integration_tests/suite/test_calls_recording.py
```diff
@@ -106,6 +105,28 @@ def event_received():
...
+        )
+
+        variables = {'XIVO_CALLRECORDFILE': ''}
+        channel_id = self.given_call_not_stasis(variables=variables)
+
+        assert_that(
+            calling(self.calld_client.calls.start_record_from_user).with_args(channel_id),
+            raises(CalldError).matching(has_properties(status_code=403))
+        )
+
```
bloom1, line 129: Zuul is dead, this probably does not pass tests

bloom1:
recheck

integration_tests/suite/test_calls_recording.py
bloom1: Well it passes tests but I don't know why. Isn't the status_code supposed to be 500 in this case?

fblackburn1:
recheck

PR #147, opened 2021-02-04
WAZO-2085: handle spaces in extensions

bloom1:


pc-m (changes requested):
Could you do the same for the transfer and application plugins?

pc-m (approved):

PR #157, opened 2021-04-13
test: do not use deprecated acls

fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-test-helpers/pull/50


fblackburn1:
recheck

PR #160, opened 2021-04-23
test: use explicit ipv4 addr to connect docker

fblackburn1:
reason: service_port() returns only port bound to ipv4. Some host can
resolve localhost to ipv6 address and docker doesn't necessarily bind
the same port number on ipv4 and ipv6


fblackburn1:
recheck

PR #161, opened 2021-05-04
conference: optimize confd calls

pc-m:
Given the same conference ID avoid doing multiple queries on wazo-confd. This is
especially useful when a paging is used. In these scenario each members of the
paging will generate 2 confd calls which will then generate 1 auth call.

For a paging with 100 members this reduces the number of HTTP requests from 400
to 2.

The cache is reset whenever a conference is added, updated or deleted

### DIFF
--- a/wazo_calld/plugin_helpers/confd.py
+++ b/wazo_calld/plugin_helpers/confd.py
@@ -1,5 +1,7 @@
-# Copyright 2016-2020 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2016-2021 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
+
+from functools import lru_cache
 
 from requests import HTTPError
 from requests import RequestException
@@ -119,15 +121,25 @@ class Conference:
 
     @classmethod
     def from_id(cls, conference_id, confd_client):
+        conference = cls._get_by_id_or_none(conference_id, confd_client)
+        if not conference:
+            raise NoSuchConferenceID(conference_id)
+
+        return cls(conference['tenant_uuid'], conference['id'], confd_client)
+
+    @classmethod
+    def reset_cache(cls, _):
+        cls._get_by_id_or_none.cache_clear()
+
+    @classmethod
+    @lru_cache(maxsize=128)
+    def _get_by_id_or_none(cls, conference_id, confd_client):
         try:
-            conference = confd_client.conferences.get(conference_id)
+            return confd_client.conferences.get(conference_id)
         except HTTPError as e:
-            if e.response is not None and e.response.status_code == 404:
-                raise NoSuchConferenceID(conference_id)
+            if not_found(e):
+                return None
             raise
-        return cls(conference['tenant_uuid'],
-                   conference['id'],
-                   confd_client)
 
 
 def get_voicemail(voicemail_id, confd_client):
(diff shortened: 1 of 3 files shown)


wazo_calld/plugin_helpers/confd.py
```diff
@@ -119,15 +121,25 @@ def exists(self):
...
+
+        return cls(conference['tenant_uuid'], conference['id'], confd_client)
+
+    @classmethod
+    def reset_cache(cls, _):
+        cls._get_by_id_or_none.cache_clear()
+
+    @classmethod
+    @lru_cache(maxsize=128)
+    def _get_by_id_or_none(cls, conference_id, confd_client):
```
pc-m, line 136: lru_cache will not cache exceptions. 404 is caught here and returned as None, all other exceptions are raised to avoid caching.

sduthil (approved):

sduthil:
I'm OK with this quick solution.
After reading some code, I see that we only need the wazo-confd call to get the tenant of the conference.
We could use the WAZO_TENANT_UUID channel variable for this. The downside is that it's the channel's tenant uuid, not the bridge, but I don't really see how to add a tenant uuid on a confbridge...

fblackburn1:
> I'm OK with this quick solution.
> After reading some code, I see that we only need the wazo-confd call to get the tenant of the conference.
> We could use the WAZO_TENANT_UUID channel variable for this, passed along in the event variables. The downside is that it's the channel's tenant uuid, not the bridge, but I don't really see how to add a tenant uuid on a confbridge...

[I didn't read the code] confbridge name already have the tenant slug, so maybe we can use it instead of the uuid?

sduthil:
That would be too kludgy: the tenant UUID will be forwarded into the Wazo bus event. We need the tenant UUID one way or another.

pc-m:
> I'm OK with this quick solution.
> After reading some code, I see that we only need the wazo-confd call to get the tenant of the conference.
> We could use the WAZO_TENANT_UUID channel variable for this, passed along in the event variables. The downside is that it's the channel's tenant uuid, not the bridge, but I don't really see how to add a tenant uuid on a confbridge...

Using the `WAZO_TENANT_UUID` would work to create the conference event without calling confd. The problem is that it's not a conference, it's a paging.

pc-m:
> > I'm OK with this quick solution.
> > After reading some code, I see that we only need the wazo-confd call to get the tenant of the conference.
> > We could use the WAZO_TENANT_UUID channel variable for this, passed along in the event variables. The downside is that it's the channel's tenant uuid, not the bridge, but I don't really see how to add a tenant uuid on a confbridge...
> 
> Using the `WAZO_TENANT_UUID` would work to create the conference event without calling confd. The problem is that it's not a conference, it's a paging.

I've got another idea, we could call `Confbridge` with a constructed name in the dialplan and the ignore events that do not match that pre-defined name in calld.

For example `Confbridge(wazo-conference-${WAZO_CONFBRIDGE_ID},,,${WAZO_CONFBRIDGE_MENU}`

That way the auto-generated name used for pagings would not match that `wazo-conference-<id>` pattern

sduthil:
I completely agree about the WAZO_TENANT_UUID variable. Thanks for pointing it out.
I kind of like the `wazo-conference-` prefix, even though a dedicated field would be better. Lacking the dedicated field, that I think that solution is the cleanest way so far.

sduthil:
Closing in favor of:
* wazo-platform/xivo-config#45
* #163

PR #167, opened 2021-05-26
add /config API allowing change of debug setting

sduthil:
Depends-On: https://github.com/wazo-platform/wazo-calld-client/pull/38

### DIFF
--- a/wazo_calld/plugins/config/http.py
+++ b/wazo_calld/plugins/config/http.py
@@ -1,0 +1,26 @@
+# Copyright 2021 The Wazo Authors  (see the AUTHORS file)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from flask import request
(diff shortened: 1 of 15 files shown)


sduthil:
recheck

wazo_calld/plugins/config/http.py
```diff
@@ -0,0 +1,26 @@
...
+from .schemas import config_patch_schema
+
+
+class ConfigResource(AuthResource):
+    def __init__(self, config_service):
+        self._config_service = config_service
+
+    @required_acl('calld.config.read')
+    def get(self):
+        return self._config_service.get_config(), 200
```
fblackburn1, line 18: Should be restricted to master tenant only, otherwise it can leaks sensitive config

fblackburn1 (approved):

PR #169, opened 2021-06-09
Wazo 2393 switchboard moh

pc-m:
Depends-On: https://github.com/wazo-platform/wazo-confd/pull/200

### DIFF
--- a/wazo_calld/plugins/switchboards/confd.py
+++ b/wazo_calld/plugins/switchboards/confd.py
@@ -17,19 +17,19 @@ class Switchboard:
 
     def exists(self):
         try:
-            self._get_queue()
+            self._get()
         except NoSuchSwitchboard:
             return False
         else:
             return True
 
     def hold_moh(self):
-        return self._get_queue()['waiting_room_music_on_hold']
+        return self._get()['waiting_room_music_on_hold']
 
     def queue_moh(self):
-        return self._get_queue()['queue_music_on_hold']
+        return self._get()['queue_music_on_hold']
 
-    def _get_queue(self):
+    def _get(self):
         try:
             return self._confd.switchboards.get(self.uuid, tenant_uuid=self.tenant_uuid)
         except HTTPError as e:
--- a/wazo_calld/plugins/switchboards/confd.py
+++ b/wazo_calld/plugins/switchboards/confd.py
@@ -1,6 +1,7 @@
-# Copyright 2017-2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2017-2021 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from wazo_calld.plugins.switchboards.exceptions import NoSuchSwitchboard
 from requests import HTTPError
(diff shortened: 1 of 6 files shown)


wazo_calld/plugins/switchboards/stasis.py
```diff
@@ -46,15 +46,20 @@ def notify_all_switchboard_held(self):
...
+                raise Exception('not enough arguments')
+
+            # app_instance = event['args'][0]
+            if event['args'][1] == 'switchboard_queue':
+                self._stasis_start_queue(event_objects, event)
+            elif event['args'][1] == 'switchboard_answer':
+                self._stasis_start_answer(event_objects, event)
+            elif event['args'][1] == 'switchboard_unhold':
+                self._stasis_start_answer_held(event_objects, event)
+        except Exception:
```
fblackburn1, line 60: Is it possible to use more specific error? RequestException or HTTPError?

wazo_calld/plugins/switchboards/confd.py
```diff
@@ -16,12 +17,24 @@ def __init__(self, tenant_uuid, uuid, confd_client):
...
+        else:
+            return True
+
+    def hold_moh(self):
+        return self._get_queue()['waiting_room_music_on_hold']
+
+    def queue_moh(self):
+        return self._get_queue()['queue_music_on_hold']
+
+    def _get_queue(self):
```
fblackburn1, line 32: I think that you refer to asterisk-queue, but to avoid confusion with wazo-queue
```suggestion
    def _get(self):
```
or
```suggestion
    def _get_switchboard(self):
```

fblackburn1 (changes requested):
I think a small integration tests (without corner case / failing scenario) would be useful and easy (if you can use same logic than test_applications.py)

wazo_calld/plugins/switchboards/stasis.py
pc-m: I want to catch pretty much everything here.

If an exception is raised during the stasis call handling the call will be stuck in stasis until Asterisk is restarted or the caller gets tired and hangup.

If anything bad happen planned or not the call should be sent back to the dialplan.

wazo_calld/plugins/switchboards/confd.py
pc-m: fixed

pc-m:
> I think a small integration tests (without corner case / failing scenario) would be useful and easy (if you can use same logic than test_applications.py)

I tried and talked about it with @sduthil. The problem is that there is no event when a MOH is started on a bridge, the event is only when the MOH starts on a channel.

To test that the MOH has started would need an asterisk patch or log parsing.

wazo_calld/plugins/switchboards/stasis.py
fblackburn1: Ok :ok_hand:

fblackburn1 (dismissed):

pc-m:
recheck

fblackburn1:
recheck

fblackburn1 (dismissed):

PR #173, opened 2021-06-23
Wazo 2353 event sent before answer from recipient

JaysLachance:


sduthil:
The current implementation causes side-effect issues and is not mergeable as-is. Will be reopened when working on this topic.

PR #174, opened 2021-06-25
Wazo 915 event call ringing

JaysLachance:


fblackburn1 (changes requested):
These changes have the same behavior than before.
Ticket need to be understood before coding

PR #175, opened 2021-06-28
Wazo 2001 direction of call on hangup

oxynor:


wazo_calld/plugins/calls/services.py
```diff
@@ -308,7 +308,8 @@ def make_call_from_stasis_event(event):
         call.talking_to = {}
         call.sip_call_id = channel_variables.get('WAZO_SIP_CALL_ID')
         call.line_id = channel_variables.get('WAZO_LINE_ID')
-        call.creation_time = channel.get('creationtime')
+        call.creation_time = channel_info.get('creationtime')
+        call.is_caller = channel_variables.get('WAZO_CHANNEL_DIRECTION')
```
fblackburn1, line 312: `is_caller` is a boolean `True`/`False` you need to use the same logic than:
https://github.com/wazo-platform/wazo-calld/blob/master/wazo_calld/plugin_helpers/ari_.py#L213-L226

Since you add  the variable in all messages, you can probably rework this helper to avoid fetching variable in a second (useless) call to ari.

fblackburn1 (changes requested):
An integration test will be necessary to validate this fix

wazo_calld/plugins/calls/services.py
oxynor: I'm a bit confused here. Since this method is now unique to channelDestroyed Event. I cannot use is_caller because it always return False (since the channel is destroyed). So what i'm doing here is getting the direction before the channel gets destroyed. Is that acceptable ? Seb told me that from-wazo/to-wazo was more descriptive than True and False too.

fblackburn1: You added `WAZO_CHANNEL_DIRECTION` available to *all* channels. Even if this method is only use in one event, you must avoid to add two ways to do the same thing (aka technical debt). But You can ignore it for now, and do it in a second time.

The `from-wazo/to-wazo` is more explicit if the variable is called `direction` but it doesn't make sens when the variable is called `is_caller`. I'm pretty sure he would talked about the `WAZO_CHANNEL_DIRECTION` @sduthil

sduthil: +1 with François: the `from-wazo/to-wazo` is made for the Asterisk variable (it's more explicit when working in the code), but the Wazo events should have `is_caller = True/False`, which is simpler when using the API/events.

What you did with getting the direction directly from the event is exactly what we need. You just need to convert it to boolean value, using the helper as a reference (not the ARI GET part, only the conversion string to boolean part).

fblackburn1 (changes requested):
I rebased your branch
I didn't validate the branch, only really quick look

* changelog is missing
* Please update your copyright when you modify files (see [git-hooks](https://github.com/wazo-platform/xivo-tools/tree/master/git-hooks))

pc-m (approved):

fblackburn1 (approved):

PR #176, opened 2021-06-30
Wazo 2347 missing hangup code

oxynor:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 21.09
+
+* Added a reason code in order to know why a call was ended
+* The event `call_missed` is now sent when the caller himself hangs up the call he made before the recipient could answer.
```
fblackburn1, line 6: This change does not belong to this branch

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 21.09
+
+* Added a reason code in order to know why a call was ended
```
fblackburn1, line 5: CHANGELOG file serves to inform API users what changes are done
To be relevant, you must specify which API it's affect and what is the new field name. No need to add description of the field, the openapi documentation will care of it

fblackburn1 (changes requested):
Changes in openapi (api.yml) are missing

wazo_calld/plugins/calls/services.py
```diff
@@ -309,6 +309,9 @@ def make_call_from_stasis_event(event):
         call.sip_call_id = channel_variables.get('WAZO_SIP_CALL_ID')
         call.line_id = channel_variables.get('WAZO_LINE_ID')
         call.creation_time = channel.get('creationtime')
+        call.is_caller = channel_variables.get('WAZO_CHANNEL_DIRECTION')
+        call.reason_code = event['cause']
```
fblackburn1, line 313: Are you sure that ALL stasis events have `cause` field? If not, an error will be raised

oxynor: Technically, this method is only used for a channel destoyed event, maybe the name of the method is too generic, is it ok if i simply change the name of the method to something more specific like 'channel_destroyed_event' ? That would imply that there's always a `cause` field since it's always part of the channelDestroyed stasis event. Thanks.

fblackburn1: Ah no it's ok

CHANGELOG.md
oxynor: Unless i missed something, this attribute only affect the `call_ended `event. I fixed it and added which attributes was added (`reason_code`) and also added it affects the `call_ended` event.

oxynor: Unless i missed something, this attribute only affect the `call_ended `event. I fixed it and added which attributes was added (`reason_code`) and also added it affects the `call_ended` event.

fblackburn1: Forget my openapi comment (last line), the implementation introduced my in error (I'll add comment on the right code section)
Your changelog need to be more explicit on which event is affected: ex: 
```
* The event `call_ended` now send `reason_code` to know why a call was ended`.
```

wazo_calld/plugins/calls/services.py
fblackburn1, line 313: `reason_code` is not a part of `Call` object attribute. It should be added only when the event is created

wazo_calld/plugins/calls/schemas.py
```diff
@@ -82,6 +82,7 @@ class CallSchema(CallBaseSchema):
     dialed_extension = fields.String()
     sip_call_id = fields.String()
     line_id = fields.Integer()
+    reason_code = fields.Integer()
```
fblackburn1, line 85: `reason_code` is not a part of `Call` object attribute. It should be added only when the event is created

wazo_calld/plugins/calls/services.py
```diff
@@ -309,6 +309,9 @@ def make_call_from_stasis_event(event):
...
+
```
fblackburn1, line 314: linter issue

fblackburn1 (changes requested):

wazo_calld/plugins/calls/services.py
```diff
@@ -309,6 +309,7 @@ def make_call_from_stasis_event(event):
         call.sip_call_id = channel_variables.get('WAZO_SIP_CALL_ID')
         call.line_id = channel_variables.get('WAZO_LINE_ID')
         call.creation_time = channel.get('creationtime')
+        call.is_caller = channel_variables.get('WAZO_CHANNEL_DIRECTION')
```
fblackburn1, line 312: I think you leave an artifact from another branch here

fblackburn1 (changes requested):
Since your have 3 commits with the exact same message, you can squash all these commits together

fblackburn1 (dismissed):
Work as expected, good job :+1: 
I'll add mergeit tag AFTER the freeze

fblackburn1 (approved):

PR #184, opened 2021-08-12
WAZO-2413: pause group queue member when DND

bloom1:
Depends-On: https://github.com/wazo-platform/wazo-auth-keys/pull/52


bloom1:
recheck

pc-m (changes requested):
This is a nice solution. One problem with the current implementation is that a user can be stuck in a Pause without knowing.

Given Alice is a member of the queue Q
Given Alice puts her phone in DnD mode
Given wazo-calld gets stopped
When Alice puts her phone off DnD mode
When wazo-calld gets started
Then Alice is still paused in the queue

pc-m (approved):

PR #193, opened 2021-09-23
Refactor user UUID extraction

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-lib-python/pull/86

### DIFF
--- a/wazo_calld/auth.py
+++ b/wazo_calld/auth.py
@@ -18,15 +18,13 @@ extract_token_id_from_query_or_header = auth_verifier.extract_token_id_from_quer
 Unauthorized = auth_verifier.Unauthorized
 
 
-def get_token_user_uuid_from_request(auth_client, token=None):
-    if not token:
-        token = request.headers.get('X-Auth-Token') or request.args.get('token')
+def get_token_user_uuid_from_request():
     try:
-        token_infos = auth_client.token.get(token)
+        user_uuid = request.user_uuid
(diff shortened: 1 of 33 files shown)


pc-m:
recheck

sduthil (approved):

wazo_calld/auth.py
```diff
@@ -18,15 +18,13 @@
 Unauthorized = auth_verifier.Unauthorized
 
 
-def get_token_user_uuid_from_request(auth_client, token=None):
-    if not token:
-        token = request.headers.get('X-Auth-Token') or request.args.get('token')
```
sduthil, line 23: This is not the default behavior of `AuthVerifier`: it will only look at the header, not the query string.
`AuthVerifier` needs to be instanciated with `extract_token_id = extract_token_id_from_query_or_header`. This will introduce a slight change in behavior: the token from query string will take precedence over the header.

sduthil (changes requested):

wazo_calld/auth.py
sduthil: Actually, the only place needing the `token=...` query string is voicemail recordings. Those places already have specific token extraction logic:
```python
@required_acl(
        'calld.voicemails.{voicemail_id}.messages.{message_id}.recording.read',
        extract_token_id=extract_token_id_from_query_or_header,
    )
```

sduthil (approved):

PR #197, opened 2021-12-02
WAZO-2626: calls dnd event: ignore unknown interface

pc-m:
When a user enables/disables it's DnD feature wazo calld pauses/unpauses it's
usersharedlines interface. If the user is not a member of any group Asterisk
return the error 'Interface not found' which results in a huge traceback in the
log file.

This error should be ignored

### DIFF
--- a/wazo_calld/plugins/calls/bus_consume.py
+++ b/wazo_calld/plugins/calls/bus_consume.py
@@ -14,6 +14,7 @@ from xivo_bus.resources.common.event import ArbitraryEvent
 
 from wazo_calld.plugin_helpers import ami
 from wazo_calld.plugin_helpers.ari_ import Channel, set_channel_id_var_sync
+from wazo_calld.plugin_helpers.exceptions import WazoAmidError
 
 from .schemas import call_schema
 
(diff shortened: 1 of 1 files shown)


wazo_calld/plugins/calls/bus_consume.py
```diff
@@ -277,7 +278,13 @@ def _users_services_dnd_updated(self, event):
...
-            ami.pause_queue_member(self.ami, interface)
-        else:
-            ami.unpause_queue_member(self.ami, interface)
+        try:
+            if enabled:
+                ami.pause_queue_member(self.ami, interface)
+            else:
+                ami.unpause_queue_member(self.ami, interface)
+        except WazoAmidError as e:
+            if e.details['original_error'] == 'Interface not found':
```
bloom1, line 287: Is there a better way to test this?

pc-m: It's hard to do a good test. The bug is that log files get filled with thousands of tracebacks but I think there are now behavior change with or without handling the error

bloom1 (approved):

PR #202, opened 2022-01-11
use docker-compose from binary instead of python

fblackburn1:
why: to have the latest docker-compose


fblackburn1:
recheck

fblackburn1:
recheck

PR #205, opened 2022-02-08
Wazo 2656 meeting limit api

pc-m:
Depends-On: https://github.com/wazo-platform/wazo-calld-client/pull/45

### DIFF
--- a/integration_tests/assets/etc/wazo-calld/conf.d/20-real-asterisk.yml
+++ b/integration_tests/assets/etc/wazo-calld/conf.d/20-real-asterisk.yml
@@ -2,3 +2,4 @@ ari:
   connection:
     username: xivo
     password: xivo
+max_meeting_participants: 4
--- a/wazo_calld/plugins/meetings/api.yml
+++ b/wazo_calld/plugins/meetings/api.yml
@@ -1,4 +1,21 @@
 paths:
+  /guests/me/meetings/{meeting_uuid}/status:
+    get:
+      summary: Get the status of a meeting
+      description: Returns the status of a meeting that should be visible to a guest.
+      parameters:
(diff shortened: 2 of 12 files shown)


bloom1:
recheck

pc-m:
recheck

wazo_calld/plugins/meetings/api.yml
```diff
@@ -1,4 +1,21 @@
 paths:
+  /guests/me/meetings/{meeting_uuid}/status:
```
bloom1, line 2: The CHANGELOG entry is missing for this new route

integration_tests/assets/etc/wazo-calld/conf.d/20-real-asterisk.yml
```diff
@@ -2,3 +2,4 @@ ari:
   connection:
     username: xivo
     password: xivo
+max_meeting_participants: 4
```
bloom1, line 5: Is that enforced somewhere else or am I missing something? 
Should that be in the wazo-confd API and defined for each meeting? I'm not sure I understand

bloom1 (changes requested):

integration_tests/assets/etc/wazo-calld/conf.d/20-real-asterisk.yml
pc-m: In `config.py` it is set to 25 which is the current "advertised" limit. I made it a configuration parameter to be able to test the limit without using 26 calls and to be able to change that limit more easily if/when we decide that 25 is not the limit anymore.

bloom1 (approved):

pc-m:
regate

PR #211, opened 2022-02-25
Add two fields in calld; one for the pick up time of a call, and another for the hang up time

nissou91:
We would like to have two fields in wazo-calld about each call:

* One called answer_time which will tells us when the call was picked up by the other end,

* The second one called hangup_time which will tell us when the call was ended.

* Write all the necessary test functions (possible scenarios), if needed.

Depends-On: https://github.com/wazo-platform/wazo-asterisk-config/pull/44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## 22.04
 
+* The `answer_time` field has been added to the following resources:
+
+  * GET `/calls`
+  * GET `/calls/<call-id>`
+  * GET `/users/me/calls`
+  
+* The `answer_time` field have been added to the following events:
+
+  * `call_created`
+  * `call_updated`
+  * `call_answered`
+  * `call_ended`
+  
 * The `ivr_extension` and `wait_time` fields have been added to the following resources:
 
   * POST `/faxes`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,14 @@
   * GET `/calls`
   * GET `/calls/<call-id>`
   * GET `/users/me/calls`
+
+* The `hangup_time` field has been added to the following resources:
+
+  * GET `/calls`
+  * GET `/calls/<call-id>`
+  * GET `/users/me/calls`
   
-* The `answer_time` field have been added to the following events:
+* The `answer_time` and `hangup_time` fields have been added to the following events:
 
   * `call_created`
   * `call_updated`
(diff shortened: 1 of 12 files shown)


pc-m (changes requested):
Hi Aniss,
There are still some missing parts in your branch.

1. If I use a `GET` on a call or a list of calls both fields should be in the result.
2. You probably also need to implement the changes in the `calls` plugin.
3. You will need to add the new fields in the api.yml file and the changelog file

To have the `pick_up_time` available during the whole lifetime of the call you will need to assign it to a channel variable in Asterisk with a line similar to this
```python
ari.channels.setChannelVar(channelId=channel_id, variable='WAZO_PICKUP_TIME', value=datetime.datetime.now())
```
and the when serializing the call object for an event or an API call you would fetch that variable back.
This is where call data get fetched https://github.com/wazo-platform/wazo-calld/blob/master/wazo_calld/plugins/calls/services.py#L265

pc-m (dismissed):
I've modified the place where the WAZO_PICKUP_TIME channel variable is set. The old location only assigned the value to the variable when using the API to answer the call. Not when answering using the phone.
I also noticed that the time format was not a date `{"value": "the date"}` instead of `"the date"`. I modified the test to add a custom date matched.
The call_ended event also need the hang_up_time. it is always None. You will not need to use a channel variable for this one since the variable is only visible once, during the call_ended event.

sduthil (changes requested):
* The name of the fields is not consistent with the rest of the code / API. AFAIK, the name of the fields has not been discussed, so they must be renamed to follow the code / API wording:
  * `answer_time`
  * `hangup_time`
* We want the new fields in all the call-related events:
  * `call_updated`
  * `call_answered`
  * `call_held`
  * `call_resumed`
  * `call_ended` (the only one having `hangup_time` not `None`)
* In order to have `answer_time` correctly set in `call_ended` event, we must make Asterisk send the channel variable along the Asterisk Stasis `ChannelDestroyed` event, because the channel will no longer exist, so we can't get the `WAZO_ANSWER_TIME`. For this, you need to add the variable `WAZO_ANSWER_TIME` in [`ari.d`](https://github.com/wazo-platform/wazo-asterisk-config/blob/master/etc/asterisk/ari.d/01-wazo.conf) (and symmetrically in [`manager.d`](https://github.com/wazo-platform/wazo-asterisk-config/blob/master/etc/asterisk/manager.d/99-general.conf). Once this is done, there will be no need for `ari.channels.getChannelVar(..., variable='WAZO_ANSWER_TIME')` because the variable will be available in `channel.json.get('channelvars', {})`.

sduthil (approved):

PR #215, opened 2022-03-14
Wazo-calld: Fix callerid parsing when it starts with at least one space, and update the appropriate tests

nissou91:
Why:

* When ```calledid``` starts with at least one space, the ```caller_name``` and ```caller_num``` values are not extracted correctly from the storage.


pc-m (changes requested):
I don't think this fixes all of the problem. If the prefix is `callerid =` the line will still be ignored. You can add the following test to check that the ticket is fixed.
```diff
--- a/wazo_calld/plugins/voicemails/tests/test_storage.py
+++ b/wazo_calld/plugins/voicemails/tests/test_storage.py
@@ -44,6 +44,37 @@ category=
 msg_id=1478200319-00000000
 flag=
 duration=12
+'''.encode('utf-8')
+        result = self._parse(content)
+        expected = {
+            'id': '1478200319-00000000',
+            'caller_id_name': 'Etienne',
+            'caller_id_num': '101',
+            'timestamp': 1478200319,
+            'duration': 12,
+        }
+        assert_that(result, equal_to(expected))
+
+    def test_parse_with_spaces(self):
+        content = '''
+;
+; Message Information file
+;
+[message]
+origmailbox = 1001
+context = user
+macrocontext = 
+exten = voicemail
+rdnis = 1001
+priority = 7
+callerchan = PJSIP/xivo64-00000000
+callerid = "Etienne" <101>
+origdate = Thu Nov  3 07:11:59 PM UTC 2016
+origtime = 1478200319
+category = 
+msg_id = 1478200319-00000000
+flag = 
+duration = 12
 '''.encode('utf-8')
         result = self._parse(content)
         expected = {
```

wazo_calld/plugins/voicemails/storage.py
```diff
@@ -243,6 +246,10 @@ def _parse_callerid(value, result):
         elif caller_id.is_complete_caller_id(value):
             result['caller_id_name'] = caller_id.extract_displayname(value)
             result['caller_id_num'] = caller_id.extract_number(value)
+        elif value != value.lstrip() and caller_id.is_complete_caller_id(value.lstrip()):
```
pc-m, line 249: I think this `elif` is not required anymore since all lines are cleaned at the top of the `for` loop.

pc-m (changes requested):

wazo_calld/plugins/voicemails/storage.py
```diff
@@ -219,6 +219,9 @@ def parse(self, fobj):
         result = {}
         parsed = set()
         for line in fobj:
+            line = '='.join([x.strip() for x in line.decode('utf-8').split('=')])
```
pc-m, line 222: Quick question. Would it work without decoding and encoding?
```python3
line = b'='.join([x.strip() for x in line.split(b'=')])
line = line.lstrip()
```

pc-m: I also think `line = line.lstrip()` is redundant since `x.strip()` already stripped the first element of the line. This line could probably be removed.

pc-m (changes requested):

wazo_calld/plugins/voicemails/storage.py
nissou91: Yes you are right, I removed the encoding and the ```lfstrip``` line :)

pc-m (approved):

PR #222, opened 2022-03-30
integration tests: use 127.0.0.1 for docker address

sduthil:
Why:

* if localhost resolves to IPv6, the wrong port will be used in
integration tests


bloom1:
regate

fblackburn1:
regate

PR #223, opened 2022-04-01
add logs when the ari thread starts and stops

pc-m:


fblackburn1:
@pc-m still relevant?

pc-m:
> @pc-m still relevant?

Might not be, I'm not ready to delete it

PR #225, opened 2022-04-06
tests: don't stop upon failure

sduthil:
Why:

* We now have logs of containers in the CI, so there is no need to keep
the container to inspect logs.
* We want to know how many tests are failing


sduthil:
recheck

PR #228, opened 2022-04-15
WAZO-2754-no-cancel-when-answering-from-a-mobile

pc-m:


wazo_calld/plugins/dial_mobile/services.py
```diff
@@ -281,3 +281,29 @@ def cancel_push_mobile(self, call_id):
             pending_push.tenant_uuid,
             pending_push.user_uuid,
         )
+
+    def remove_pending_push_mobile(self, call_id):
+        self._pending_push_mobile.pop(call_id, None)
+
+    def is_push_answered_by_mobile(self, push_call_id, call_id, endpoint, user_uuid):
```
sduthil, line 288: The name `answer` is misleading, because it does not need to be the mobile that answers the caller, but it only needs the mobile to have registered after a push notification.

Suggestion: `is_push_received_by_mobile` or `has_pending_push`+ `is_mobile_registered`

sduthil (changes requested):

wazo_calld/plugins/dial_mobile/bus_consume.py
```diff
@@ -54,10 +56,42 @@ def _on_refresh_token_deleted(self, event):
         self._service.on_mobile_refresh_token_deleted(event['user_uuid'])
 
     def _on_bridge_enter(self, event):
-        self._service.cancel_push_mobile(event['Uniqueid'])
+        if event['BridgeType'] != 'stasis':
+            # Ignoring unrelated bridges
+            return
```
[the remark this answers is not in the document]
pc-m: > Why are Stasis bridges ignored? CTI transfers are made via `stasis` bridges.

All bridges created by the `dial_mobile` application are of type `stasis`. I filter out all none `stasis` to avoid doing ARI HTTP requests later for unrelated bridges.

wazo_calld/plugins/dial_mobile/services.py
pc-m: This is only for answered calls. We check if the device that entered the bridge is registered as mobile and owned by the user being called.

sduthil (approved):

PR #232, opened 2022-05-26
WAZO-2096: add call direction

bloom1:
Depends-On: https://github.com/wazo-platform/xivo-config/pull/66
Depends-On: https://github.com/wazo-platform/wazo-asterisk-config/pull/47

### DIFF
--- a/integration_tests/suite/test_calls.py
+++ b/integration_tests/suite/test_calls.py
@@ -131,96 +131,11 @@ class TestListCalls(IntegrationTest):
             )
         )))
 
-    def test_call_direction_internal_internal(self):
+    def test_call_direction(self):
         self.ari.set_channels(MockChannel(id='first-id'), MockChannel(id='second-id'))
         self.ari.set_channel_variable(
             {
                 'first-id': {'WAZO_CALL_DIRECTION': 'internal'},
-                'second-id': {'WAZO_CALL_DIRECTION': 'internal'},
-            }
-        )
-        self.ari.set_bridges(MockBridge(id='bridge-id', channels=['first-id', 'second-id']))
-
-        calls = self.calld_client.calls.list_calls()
-
-        assert_that(calls, has_entries(items=contains_inanyorder(
-            has_entries(call_id='first-id', direction='internal'),
-            has_entries(call_id='second-id', direction='internal'),
-        )))
-
-    def test_call_direction_internal_outbound(self):
-        self.ari.set_channels(MockChannel(id='first-id'), MockChannel(id='second-id'))
-        self.ari.set_channel_variable(
-            {
-                'first-id': {'WAZO_CALL_DIRECTION': 'internal'},
-                'second-id': {'WAZO_CALL_DIRECTION': 'outbound'},
-            }
-        )
-        self.ari.set_bridges(MockBridge(id='bridge-id', channels=['first-id', 'second-id']))
-
-        calls = self.calld_client.calls.list_calls()
-
-        assert_that(calls, has_entries(items=contains_inanyorder(
-            has_entries(call_id='first-id', direction='outbound'),
-            has_entries(call_id='second-id', direction='outbound'),
-        )))
-
-    def test_call_direction_internal_inbound(self):
-        self.ari.set_channels(MockChannel(id='first-id'), MockChannel(id='second-id'))
-        self.ari.set_channel_variable(
-            {
-                'first-id': {'WAZO_CALL_DIRECTION': 'inbound'},
-                'second-id': {'WAZO_CALL_DIRECTION': 'internal'},
-            }
-        )
-        self.ari.set_bridges(MockBridge(id='bridge-id', channels=['first-id', 'second-id']))
-
-        calls = self.calld_client.calls.list_calls()
-
-        assert_that(calls, has_entries(items=contains_inanyorder(
-            has_entries(call_id='first-id', direction='inbound'),
-            has_entries(call_id='second-id', direction='inbound'),
-        )))
-
-    def test_call_direction_inbound_inbound(self):
-        self.ari.set_channels(MockChannel(id='first-id'), MockChannel(id='second-id'))
-        self.ari.set_channel_variable(
-            {
-                'first-id': {'WAZO_CALL_DIRECTION': 'inbound'},
-                'second-id': {'WAZO_CALL_DIRECTION': 'inbound'},
-            }
-        )
-        self.ari.set_bridges(MockBridge(id='bridge-id', channels=['first-id', 'second-id']))
-
-        calls = self.calld_client.calls.list_calls()
-
-        assert_that(calls, has_entries(items=contains_inanyorder(
-            has_entries(call_id='first-id', direction='inbound'),
-            has_entries(call_id='second-id', direction='inbound'),
-        )))
-
-    def test_call_direction_inbound_outbound(self):
-        self.ari.set_channels(MockChannel(id='first-id'), MockChannel(id='second-id'))
-        self.ari.set_channel_variable(
-            {
-                'first-id': {'WAZO_CALL_DIRECTION': 'inbound'},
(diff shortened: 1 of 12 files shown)


bloom1:
recheck

integration_tests/suite/test_calls_bus_consume.py
```diff
@@ -104,7 +107,152 @@ def assert_function():
...
+                    has_entries(
+                        message=has_entries({
+                            'name': 'call_updated',
+                            'origin_uuid': XIVO_UUID,
+                            'data': has_entries({
+                                'call_id': first_channel_id,
+                                'status': 'Up',
+                                'hangup_time': None,
+                                'answer_time': is_(a_timestamp()),
+                                'is_video': True,
```
sduthil, line 173: We don't care about `answer_time` or `is_video` in this test. Please reduce the test to the minimum :)

integration_tests/suite/test_calls_bus_consume.py
```diff
@@ -104,7 +107,152 @@ def assert_function():
...
```
sduthil, line 254: Idem

integration_tests/suite/test_calls_bus_consume.py
```diff
@@ -104,7 +107,152 @@ def assert_function():
...
+                                'direction': 'outbound',
```
sduthil, line 255: In those tests, the value of `direction` did not change from its original value. I think another test would be also interesting: 3 channels inbound + internal + internal -> inbound leaves -> calls become `direction: internal`, or something like that, just to test that the `direction` can dynamically change its value during a call.

wazo_calld/plugins/calls/bus_consume.py
```diff
@@ -210,13 +210,21 @@ def _relay_channel_entered_bridge(self, event):
                 logger.debug('channel %s not found', participant_channel_id)
                 return
 
+            set_channel_id_var_sync(
+                self.ari,
+                participant_channel_id,
+                'WAZO_CONVERSATION_DIRECTION',
+                '',  # NOTE(afournier): it will be recalculated in `make_call_from_channel`
+                bypass_stasis=True,
+            )
```
sduthil, line 219: Maybe extract this to a function `invalidate_conversation_direction_cache` or equivalent?

wazo_calld/plugins/calls/bus_consume.py
```diff
@@ -235,6 +243,15 @@ def _relay_channel_left_bridge(self, event):
                 logger.debug('channel %s not found', participant_channel_id)
                 return
 
+            if channels_in_bridge > 1:
+                # If this is the last channel of the bridge we want to keep the last call direction
```
sduthil, line 247: Please reformulate this comment: the condition in the comment is the reverse of the condition in the code, so I had a hard time understanding the comment.

wazo_calld/plugins/calls/services.py
```diff
@@ -492,6 +498,37 @@ def set_answered_time(self, channel_id):
             logger.debug('channel %s not found', channel_id)
             raise NoSuchCall(channel_id)
 
+    @staticmethod
```
sduthil, line 501: Since this function needs `ari`, what's the advantage to making it a `staticmethod`, instead of using `self._ari`?

wazo_calld/plugins/calls/services.py
```diff
@@ -492,6 +498,37 @@ def set_answered_time(self, channel_id):
...
+        if 'outbound' in directions and 'inbound' in directions:
+            return 'unknown'
+
+        if 'outbound' in directions:
+            return 'outbound'
+
+        if 'inbound' in directions:
+            return 'inbound'
+
+        # NOTE(afournier): internal should have the lowest priority
```
sduthil, line 529: I don't think this comment very relevant. I think the code is explicit enough. If you have a specific reason for the comment, please add it. Else, I think we can remove this comment.

wazo_calld/plugins/calls/tests/test_services.py
```diff
@@ -104,3 +106,42 @@ def test_direction_of_call_to_who_is_caller(self):
...
+        )
+
+    def test_call_direction(self):
+
+        inbound_channel = 'inbound'
+        outbound_channel = 'outbound'
+        internal_channel = 'internal'
+        unknown_channel = 'unknown'
+
+        self._assert_conversation_direction([], internal_channel)
```
sduthil, line 123: Suggestion: I would find this clearer to read:
```python
direction = self.services._conversation_direction_from_directions
assert_that(direction([internal, internal]), equal_to(internal))
assert_that(direction([inbound, internal]), equal_to(inbound))
...
```

integration_tests/suite/test_calls.py
```diff
@@ -131,6 +131,108 @@ def test_given_some_calls_with_user_id_when_list_calls_then_calls_are_complete(s
...
+            }
+        )
+        self.ari.set_bridges(MockBridge(id='bridge-id', channels=['first-id', 'second-id']))
+
+        calls = self.calld_client.calls.list_calls()
+
+        assert_that(calls, has_entries(items=contains_inanyorder(
+            has_entries(call_id='first-id', direction='outbound'),
+            has_entries(call_id='second-id', direction='outbound'),
+        )))
```
sduthil, line 166: All those extra scenarios are redundant with the unit tests that compute the direction has the correct value. I think only one integration test would be enough.

sduthil (changes requested):

wazo_calld/plugins/calls/bus_consume.py
bloom1: I will only remove this comment, I now find it pretty useless.

wazo_calld/plugins/calls/services.py
bloom1: I will remove it.

wazo_calld/plugins/calls/tests/test_services.py
bloom1: I agree. Done :smile:

wazo_calld/plugins/calls/services.py
bloom1: That's because it is called through `make_call_from_channel` (`calls/services.py:294`) which is also a `staticmethod`. I cannot change it easily without touching multiple other plugins so I avoided changing it to a standard method.

sduthil: Oh yes, of course :) Thanks for the explanation.

bloom1:
recheck

bloom1:
> Build failed.
> 
>     * [tox-linters ](https://zuul.wazo.community/zuul/t/local/build/029b032e33394208a6942b63914a5eea) : SUCCESS in 4m 17s
> 
>     * [wazo-tox-py37 ](https://zuul.wazo.community/zuul/t/local/build/116b36ae6a4e40a3bfe4f0dc24d76cf6) : SUCCESS in 4m 47s
> 
>     * [debian-packaging ](https://zuul.wazo.community/zuul/t/local/build/040ff37d19e2490180c115c4c57457b2) : SUCCESS in 5m 13s
> 
>     * [calld-tox-integration ](https://zuul.wazo.community/zuul/t/local/build/88f67a3ef8194333933f7889defdd504) : FAILURE in 1h 00m 36s

The failing test passes with multiple individual runs and with running the whole test file multiple times on my machine :confused:

bloom1:
recheck

sduthil (changes requested):
I think we can do better about performance: we should only update the `WAZO_CONVERSATION_DIRECTION` cache when the value changes. For example, it cannot change on a `wazo_call_answered` event. In other cases, we can determine if the value changed (by comparing the value from the event and the newly computed value) and only update the cache when necessary.

bloom1:
recheck

pc-m (approved):

pc-m:
recheck

sduthil:
recheck

pc-m:
regate

pc-m:
regate

PR #242, opened 2022-10-03
pin importlib_metadata package

fblackburn1:
why: kombu use this library without pinning, and the latest version
(5.0.0) of importlib_metadata break the API.
We pin the library to the same version of current (buster) debian version


fblackburn1 (commented):
false positive

PR #243, opened 2022-11-04
mock: check responses and raise on error

pc-m:
This change will not change anything. The goal is to make HTTP errors raise exception during tests instead of failing silently.
When trying to figure out what happened I need to look at the logs trying to find the matching timestamp to view the query result from the remote service to know the result.


integration_tests/suite/helpers/phoned.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2022 The Wazo Authors  (see the AUTHORS file)
```
fblackburn1, line 1:
```suggestion
# Copyright 2020-2022 The Wazo Authors  (see the AUTHORS file)
```

sduthil (approved):

pc-m:
regate

fblackburn1:
regate

fblackburn1:
regate

PR #254, opened 2023-02-28
requirements: use digium release for swagger-py

fblackburn1:
why: A commit to support Python 3 has been accepted in upstream repo in
2019. But no version has been released


fblackburn1:
Too bad ...
I won't have time now to investigate

sopelj:
> Too bad ... I won't have time now to investigate

It looks like they merged some fixes for Python3 into master, but clearly there was a merge issue and in master there is still a Python2 import right below the try catch for dual compatibility:

https://github.com/digium/swagger-py/blob/6e7f6b9a0558b5220d205ff00de731f1fef252e8/swaggerpy/swagger_model.py#L14

So, we can't import from that file without patching it.

sduthil:
recheck

PR #255, opened 2023-03-03
test_calls_stasis: wait for bus before starting the test

sduthil:
Why:

* Now that stasis uses the bus, we need to wait for the bus


fblackburn1:
recheck (failing tests are not related to this fix)

PR #260, opened 2023-03-24
Add lock to avoid multiple answers on held calls on switchboards

xxbillouxx:


xxbillouxx:
recheck

xxbillouxx:
recheck

xxbillouxx:
recheck

xxbillouxx:
recheck

fblackburn1:
@xxbillouxx it's not normal to trigger recheck when a test added in the same PR fail 75% of the time

PR #261, opened 2023-03-27
stasis: fix hangup channel alone

sduthil:
Why:

* In some cases, the left channel may not be hungup yet (because of
  Asterisk hangup processing), but we still want to hang up the other
  party.


sduthil:
Wrong branch name

PR #262, opened 2023-03-27
stasis: fix hangup channel alone

sduthil:
Why:

* In some cases, the left channel may not be hungup yet (because of
  Asterisk hangup processing), but we still want to hang up the other
  party.

### DIFF
--- a/wazo_calld/plugins/transfers/stasis.py
+++ b/wazo_calld/plugins/transfers/stasis.py
@@ -280,15 +280,9 @@ class TransfersStasis:
             return
 
         logger.debug('cleaning bridge %s', bridge.id)
-        try:
-            self.ari.channels.get(channelId=channel.id)
-            channel_is_hungup = False
-        except ARINotFound:
-            logger.debug('channel who left was hungup')
-            channel_is_hungup = True
 
-        if len(bridge.json['channels']) == 1 and channel_is_hungup:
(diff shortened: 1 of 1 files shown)


wazo_calld/plugins/transfers/stasis.py
```diff
@@ -282,13 +282,11 @@ def clean_bridge(self, channel, event):
         logger.debug('cleaning bridge %s', bridge.id)
         try:
             self.ari.channels.get(channelId=channel.id)
```
pc-m, line 284: channels.get can be removed since we are not doing anything with the result anymore

pc-m (changes requested):

wazo_calld/plugins/transfers/stasis.py
bloom1: @sduthil Do you think that this comment should be addressed now?

sduthil: Yes, thank you for the ping. It's fixed now.

sduthil:
recheck

bloom1:
recheck

pc-m (approved):

pc-m:
recheck

bloom1:
recheck

pc-m:
recheck

PR #270, opened 2023-04-18
integration tests: mount docker volumes read-only

sduthil:
Why:

* Avoid creation of __pycache__ as root user
* This makes CI fail when removing test files


sduthil:
recheck

PR #282, opened 2023-06-15
status: wait for asterisk to be ready before returning OK

pc-m:
this code will block until Asterisk is ready and return a 500 if asterisk is down it should not be merged as is


pc-m:
I implemented a WaitStrategy for tests in another PR

PR #284, opened 2023-06-22
transfers: fix transfer failure when processing is slower

sduthil:
Why:

* If event processing is slower, the two stasis channels may not get
  moved to the new bridge before they are hungup by the bridge cleaner.

### DIFF
--- a/wazo_calld/plugin_helpers/ari_.py
+++ b/wazo_calld/plugin_helpers/ari_.py
@@ -184,6 +184,12 @@ class Channel:
             return None
         return linkedid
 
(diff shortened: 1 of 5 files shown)


fblackburn1 (approved):

fblackburn1:
@sduthil I fixed import, but no explicit test make this code triggered
I won't approve in case you find a quick way to add test
Otherwise, tell me, and I will approve

fblackburn1 (dismissed):

fblackburn1 (approved):

PR #287, opened 2023-08-04
test: wait for wazo-amid to be connected to ami

fblackburn1:
why: if we start tests when wazo-amid is not connected to asterisk,
messages won't be forwarded by this one


pc-m (dismissed):

fblackburn1:
Well I guess I will need to improve the patch next week :cry:

pc-m (approved):

PR #288, opened 2023-08-08
Pull request for endpoints-plugin-status

fblackburn1:
## status: add voicemail plugin "status" field

why: according convention, every component should have a 'status' field
also add missing openapi spec definition about voicemail component

## status: add endpoints plugin status

why: callback can be long to be executed and can cause some race
condition with tests

## test: do not raise 404 with ari mock

why: because real asterisk never trigger 404 when variable doesn't exist
and this will avoid a lot of unnecessary traceback during test with
XIVO_TRANSFERS_INDEX variable


sduthil:
recheck

pc-m (approved):

PR #290, opened 2023-08-10
WAZO-3369 move-channels-with-ari-adhoc-conference

fblackburn1:
## test adhoc conference fix styling issues

why: remove 200 lines from files, easier to read

## adhoc conference: use ari to move channels to app

why: using ari allow us to have more control over bridge/channel.
By using ari, we could delete bridge without issue and avoid a race
condition that occur when cleanup from callcontrol remove bridge with
only one channel

Moreover, using ari is the future to go and using dual logic (ari when
stasis and ami when non stasis channel) make logic cleaner and standard
with other usage of redirect in the code (relocate, transfers)

Note: integration already test stasis AND non stasis channels

### DIFF
--- a/integration_tests/suite/test_adhoc_conferences.py
+++ b/integration_tests/suite/test_adhoc_conferences.py
@@ -12,6 +12,7 @@ from hamcrest import (
     has_entries,
     has_item,
     has_items,
+    has_key,
     has_length,
     has_properties,
 )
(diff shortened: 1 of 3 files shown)


fblackburn1:
recheck ... (we still have flaky tests :weary: )

sduthil (approved):

PR #291, opened 2023-08-11
DO NOT MERGE: test test-helpers-PR

sopelj:
Depends-on: https://github.com/wazo-platform/wazo-test-helpers/pull/89


fblackburn1:
recheck (the last commit has not been tested, and fail locally)

fblackburn1:
recheck

fblackburn1:
@sopelj I had the same error locally, probably that `R =` need to be outside of the if typing, because it used by `cast(R, ...)`

sopelj:
> @sopelj I had the same error locally, I think that `R =` need to be outside of the `if typing`, because it used by `cast(R, ...)`

Oh sorry. You are right. I'll fix it.

sopelj:
recheck (changes made to wazo-test-helpers PR)

PR #295, opened 2023-09-01
call: remove tracebacks when dialing a provisioning code

pc-m:
the autoprov PJSIP endpoint does not contain a tenant UUID. adding it at template creation would not fix the problem for all installed systems. Adding to confgend would be possible but it would add dependencies in wazo-confgend.

WAZO-3384

### DIFF
--- a/wazo_calld/plugins/calls/services.py
+++ b/wazo_calld/plugins/calls/services.py
@@ -34,6 +34,7 @@ CALL_RECORDING_FILENAME_TEMPLATE = (
     '/var/lib/wazo/sounds/tenants/{tenant_uuid}/monitor/{recording_uuid}.wav'
 )
 LOCAL_TIMEZONE = datetime.datetime.now(datetime.timezone.utc).astimezone().tzinfo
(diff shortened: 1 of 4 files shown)


wazo_calld/plugins/calls/services.py
```diff
@@ -443,6 +445,7 @@ def channel_destroyed_event(ari, event):
         call.line_id = channel_variables.get('WAZO_LINE_ID') or None
         call.creation_time = channel.get('creationtime')
         call.answer_time = channel_variables.get('WAZO_ANSWER_TIME') or None
+        call.is_autoprov = event['channel']['dialplan']['context'] == AUTOPROV_CONTEXT
```
ja0x, line 448: For good styling, shouldn't this attribute also be added to Call helper class in calls/call.py?

ja0x (changes requested):

wazo_calld/plugins/calls/services.py
pc-m: yep, done

ja0x (approved):

PR #301, opened 2024-01-10
Wazo 3342 editable install

fblackburn1:


fblackburn1:
recheck

fblackburn1:
Since it was a PR created for @DrPyser, I would let him the failure investigation :smile:

fblackburn1:
recheck (error was due to issue in master, now fixed)

PR #305, opened 2024-01-24
Wazo 3259 connect user timeout

DrPyser:
https://wazo-dev.atlassian.net/browse/WAZO-3259

Depends-On: https://github.com/wazo-platform/wazo-calld-client/pull/55

### DIFF
--- a/integration_tests/suite/test_calls.py
+++ b/integration_tests/suite/test_calls.py
@@ -4,6 +4,7 @@
 import json
 import uuid
 
+import pytest
 from hamcrest import (
     assert_that,
     calling,
@@ -2303,7 +2304,7 @@ class TestConnectUser(IntegrationTest):
         )
         self.ari.set_originates(MockChannel(id=my_new_call_id))
 
-        new_call = self.calld.connect_user(call_id, 'user-uuid')
+        new_call = self.calld_client.calls.connect_user(call_id, 'user-uuid')
 
         assert_that(new_call, has_entries(call_id=my_new_call_id))
 
@@ -2349,7 +2350,7 @@ class TestConnectUser(IntegrationTest):
         )
         self.ari.set_originates(MockChannel(id=my_new_call_id))
 
-        new_call = self.calld.connect_user(call_id, 'user-uuid', timeout=1)
+        new_call = self.calld_client.calls.connect_user(call_id, 'user-uuid', timeout=1)
 
         assert_that(new_call, has_entries(call_id=my_new_call_id))
         assert_that(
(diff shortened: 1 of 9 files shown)


DrPyser:
recheck

DrPyser:
depends-on: https://github.com/wazo-platform/wazo-calld-client/pull/55

DrPyser:
recheck

integration_tests/suite/test_calls.py
```diff
@@ -2305,6 +2305,52 @@ def test_given_one_call_and_one_user_when_connect_user_then_the_two_are_talking(
...
+    def test_one_call_one_user_with_timeout(
+        self,
+    ):
+        call_id = new_call_id()
+        my_new_call_id = new_call_id(leap=1)
+        self.ari.set_channels(
+            MockChannel(id=call_id),
+            MockChannel(id=my_new_call_id),
+        )
+        self.ari.set_channel_variable({my_new_call_id: {'XIVO_USERUUID': 'user-uuid'}})
```
sduthil, line 2338: Please use `WAZO_USERUUID` where possible. `XIVO_USERUUID` is deprecated.

wazo_calld/plugins/calls/http.py
```diff
@@ -274,11 +276,25 @@ def put(self, call_id):
...
+            body = connect_call_request_body_schema.load(request.json)
+
+        args = {
+            'call_id': call_id,
+            'user_uuid': user_uuid,
+        }
+
+        if body:
+            # NOTE(clanglois): setting the 'timeout' to null
+            # allows disabling the timeout(infinite ringing)
```
sduthil, line 294: This important info should go in the documentation, not in a comment.

wazo_calld/plugins/calls/services.py
```diff
@@ -328,7 +328,7 @@ def hangup_user(self, call_id, user_uuid):
         self._verify_user(call_id, user_uuid)
         self._ari.channels.hangup(channelId=call_id)
 
-    def connect_user(self, call_id, user_uuid):
+    def connect_user(self, call_id, user_uuid, timeout=30):
```
sduthil, line 331: The default value of 30 seconds should only be defined once. It is currently defined twice: here and in the schema.

sduthil (changes requested):
Thanks Charles!

A few other things are missing:
* update the API doc
* update the API changelog
* Test when timeout is invalid

wazo_calld/plugins/calls/http.py
DrPyser: why not both?
Agreed, I'll add it to the openapi doc.

sduthil (approved):

sduthil (approved):

PR #307, opened 2024-03-06
WP-1202 stereo recording

DrPyser:
mixmonitor command parameter support(through WAZO_MIXMONITOR_COMMAND variable)


pc-m:
A plugin with the patch has been published instead

PR #310, opened 2024-03-29
Pull request for WAZO-3690-transfer-race-condition

sduthil:
## transfers: split transfer object creation from transfer initiation

Why:

* Scenario:
  * transfer is initiated
  * channels are moved to stasis by the API thread
  * channels both enter stasis into the bus thread
  * the transfer object is not created yet in the API thread, preventing
    the transfer to start
  * the transfer is created too late by the API thread

## transfers: rename non-stasis states

Why:

* clarify the state we're in

## transfers: move event code to the service

Why:

* remove duplication

## transfers: move creation logic to service

Why:

* avoid duplication between TransferStateReady and
  TransferStateNonStasis

### DIFF
--- a/integration_tests/suite/test_transfers.py
+++ b/integration_tests/suite/test_transfers.py
@@ -701,9 +701,12 @@ class TestUserListTransfers(TestTransfers):
         user_uuid = 'user-uuid'
         self.auth.set_token(MockUserToken(token, user_uuid=user_uuid))
 
-        result = self.calld.list_my_transfers(token)
+        def list_is_empty():
+            result = self.calld.list_my_transfers(token)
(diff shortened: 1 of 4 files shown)


pc-m (changes requested):
Good job. There's a fixup commit left in the history. I'm pretty sure you meant to rebase/fixup that commit.

sduthil:
Tests failure were due to an unrelated flaky test. The flakiness should be fixed in this other PR (auto-merge): https://github.com/wazo-platform/wazo-calld/pull/311

sduthil:
recheck

pc-m (approved):

PR #311, opened 2024-04-03
tests pickup: remove flakiness

sduthil:
Why:

* Sometimes, the callee channel ID is the same as the caller channel ID
  due to a race condition


fblackburn1:
hmm ... `FAILED suite/test_calls.py::TestPickup::test_pickup_hangup`
I won't force regate

PR #312, opened 2024-04-03
switchboards: fix race condition raising HTTP error

sduthil:
Why:

* Error flow:
  * The HTTP request comes in to answer the queued call
  * The call to the operator is originated
  * The operator answers automatically
  * The operator call enters stasis and the two calls are bridged
  * wazo-call tries to remove the call from the queued bridge, but
    fails, since the caller channel has already been moved to another
    bridge

Depends-On: https://github.com/wazo-platform/ari-py/pull/5

PRE-MERGE STEPS:
* merge ari-py PR
* tag ari-py
* update requirements.txt

### DIFF
--- a/integration_tests/test-requirements.txt
+++ b/integration_tests/test-requirements.txt
@@ -1,5 +1,5 @@
 https://github.com/rsichny/swagger-py/archive/master.zip  # the digium release does not support python3
-https://github.com/wazo-platform/ari-py/archive/0.4.0.zip
+https://github.com/wazo-platform/ari-py/archive/master.zip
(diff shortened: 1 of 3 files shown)


pc-m (approved):

sduthil:
* applications test failure was a real error, fixed in the new patchset
*
* pickup test failure is a flaky test addressed in #311

sduthil:
I tagged / rebuilt the ari-py package, and added the `mergeit` tag here.
Tagging with auto-merge to avoid re-approving.

PR #318, opened 2024-04-25
Pull request for homogenize-stasis-start-logging

sduthil:
## reduce processing time and logging of StasisStart events

### DIFF
--- a/wazo_calld/ari_.py
+++ b/wazo_calld/ari_.py
@@ -5,7 +5,7 @@ import errno
 import logging
 import threading
 import time
-import urllib
+import urllib.parse
 from contextlib import contextmanager
 
 import ari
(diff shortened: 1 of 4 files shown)


wazo_calld/ari_.py
```diff
@@ -244,13 +244,21 @@ def init_client(self):
 
     def _subscribe_to_bus_events(self):
         for event_name in ALL_STASIS_EVENTS:
+            self._bus_consumer.subscribe(
```
pc-m, line 247: Should we check if debug is enabled before subscribing

sduthil: Good idea! I'll do that

sduthil: Actually, since the debug setting can be changed on-the-fly, it would be harder to implement, and the usefulness reduced on production systems. I propose to leave the code as-is, and leave the decision to log to the logging system.

pc-m (approved):

PR #322, opened 2024-06-14
dial_mobile: avoid getting stuck in Stasis

pc-m:
if a call tries to join a future bridge that does not exist or that has already been "consumed" by another call we hangup to avoid the call from being stuck in Stasis doing nothing


pc-m:
recheck

ja0x (approved):
LGTM.  Seems to work, but limited setup to test

PR #323, opened 2024-06-18
Pull request for WAZO-3690-transfer-race-condition-fix-3

sduthil:
## transfer states: add thread synchronization lock

Why:

* In some circumstances, the API thread is not finished with creating
  the transfer, when the Stasis thread already receives events for the
  channels creating the transfer, creating weird race conditions.


pc-m (dismissed):

sduthil:
Transfers still have a race condition issue :( Removing mergeit

pc-m (approved):

PR #337, opened 2024-08-05
WAZO-3779: creation_time in call_push_notification events

DrPyser:
depends-on: https://github.com/wazo-platform/wazo-agid/pull/161


sduthil (changes requested):

wazo_calld/plugins/dial_mobile/bus_consume.py
```diff
@@ -50,6 +60,7 @@ def _on_user_event(self, event):
             video_enabled,
             ring_timeout,
             event["Linkedid"],
+            creation_time=timestamp,
```
sduthil, line 63: I find the key name `creation_time` not very explicit, did you discuss it already?

wazo_calld/plugins/dial_mobile/services.py
```diff
@@ -351,6 +352,7 @@ def send_push_notification(
             'video': video_enabled,
             'ring_timeout': ring_timeout,
             'sip_call_id': sip_call_id,
+            'creation_time': creation_time,
```
sduthil, line 355: I find the key name `creation_time` not very explicit, did you discuss it already?

DrPyser: No. At first it was to reflect the `creation_time` field in calld call events. But I agree it could be renamed to avoid confusion with the actual call `creation_time`.

If we were to be explicit and accurate, though, it would be something like `push_mobile_time`? It would have to be distinguished from the `notification_time` generated by wazo-webhookd.

PR #339, opened 2024-08-09
Pull request for typing

sduthil:
## plugins: add typing to plugin.py

### DIFF
--- a/integration_tests/assets/ari_data/mock_ari.py
+++ b/integration_tests/assets/ari_data/mock_ari.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # Copyright 2015-2024 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
+
 from __future__ import annotations
 
 import json
--- a/integration_tests/assets/ari_data/mock_ari.py
+++ b/integration_tests/assets/ari_data/mock_ari.py
@@ -6,14 +6,10 @@ from __future__ import annotations
 import json
 import logging
 import sys
-from typing import TYPE_CHECKING
 
 from flask import Flask, Response, jsonify, make_response, request
 from flask_sockets import Sockets
-
-if TYPE_CHECKING:
-    from geventwebsocket.websocket import WebSocket
-
+from geventwebsocket.websocket import WebSocket
 
 _EMPTY_RESPONSES: dict[str, dict | list] = {
     'amqp': {},
--- a/wazo_calld/config.py
+++ b/wazo_calld/config.py
@@ -1,15 +1,20 @@
 # Copyright 2015-2024 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from __future__ import annotations
 
 import argparse
+from collections.abc import Sequence
+from typing import Any
 
 from xivo.chain_map import ChainMap
 from xivo.config_helper import parse_config_file, read_config_file_hierarchy
 from xivo.xivo_logging import get_log_level_by_name
 
+from wazo_calld.types import CalldConfigDict
+
 _DEFAULT_HTTP_PORT = 9500
-_DEFAULT_CONFIG = {
+_DEFAULT_CONFIG: CalldConfigDict = {
     'config_file': '/etc/wazo-calld/config.yml',
     'extra_config_files': '/etc/wazo-calld/conf.d/',
     'debug': False,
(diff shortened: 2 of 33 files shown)


wazo_calld/config.py
```diff
@@ -1,15 +1,23 @@
...
 
 from xivo.chain_map import ChainMap
 from xivo.config_helper import parse_config_file, read_config_file_hierarchy
 from xivo.xivo_logging import get_log_level_by_name
 
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from typing import Any
+
+    from wazo_calld.types import CalldConfigDict
```
fblackburn1, line 17: @DrPyser Out of curiosity, do you know if it's a good practice to make every typing import in the TYPE_CHECKING?
From what I understand, it should only be used on import conflict:
- collections: built-in library, so not useful
- typing: already imported by TYPE_CHECKING, so not useful
- wazo_calld.type: already in the same module than this code, so not useful

wazo_calld/config.py
```diff
@@ -1,15 +1,23 @@
 # Copyright 2015-2024 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+from __future__ import annotations
```
fblackburn1, line 4: From all our repos, on 739 import `__future__`, there are 519 imports without blank line before license comment

wazo_calld/config.py
sduthil: My thinking was:
1. indicate that those imports are *only* for typing
2. reduce imports to the strict minimum when running production code

wazo_calld/config.py
sduthil: Yep, I saw both forms, so I chose the one I preferred...
I find it so ugly without the space, objection to changes the 519 without spaces?

sduthil: FYI CPython adds a blank line: https://github.com/python/cpython/blob/a07cf4ce25205d836a6b30cb629eda139d2a518e/Doc/tools/extensions/c_annotations.py#L12

fblackburn1: no objection

wazo_calld/config.py
fblackburn1: Yep totally agree with you
But it's not the strategy used for other typed modules.
ex: https://github.com/python/cpython/blob/a07cf4ce25205d836a6b30cb629eda139d2a518e/Doc/tools/extensions/patchlevel.py#L6

fblackburn1: Well after search, it seems useless to add `from typing` in this block
But other imports seem to make sense
For the self types module ... I'm note sure, but I can live with it :smile:

pc-m (commented):
Looks good to a typing neophyte

integration_tests/assets/ari_data/mock_ari.py
```diff
@@ -28,7 +28,7 @@
 
 app = Flask(__name__)
 sockets = Sockets(app)
-websocket: WebSocket = None  # type: ignore[assignment]
+websocket: WebSocket = None
```
DrPyser, line 31: Why remove the type ignore? Doesn't mypy complain that you're assigning `None` to a non-optional type?
Alternatively, annotate with `WebSocket | None`?

sduthil: well, mypy did not complain after removing this, so I dropped it

DrPyser: Weird, the annotation is obviously wrong in this case, unless there's a setting to support implicit optional(which is a bad practice and not the case in other wazo projects)

DrPyser: Likely mypy is ignoring something because it doesn't understand the type information from the `geventwebsocket.websocket` import(because of the `ignore_missing_import` setting, if there's no type stub for `geventwebsocket.websocket`)

wazo_calld/config.py
DrPyser: Otherwise I won't object in this case, I just wouldn't care to lobby for this pattern everywhere.

DrPyser (commented):
Great work, that's some healthy servings of typing!

integration_tests/assets/ari_data/mock_ari.py
sduthil: OK thanks, I'll revert it back then

wazo_calld/config.py
sduthil: Thanks for your feedback, I'll remove the conditional import unless strictly necessary then.

integration_tests/assets/ari_data/mock_ari.py
DrPyser: In general though, what I would prefer for this kind of pattern is to have the runtime code using that variable explicitly acknowledge the expectation that the value is not `None`, by using `assert websocket is not None` for example. This would silence mypy selectively when conscious effort is made by the developer, but mypy would still remind someone who forgot that the variable might be None.

In this case, I don't think anything prevents `websocket` from being None. it's just a matter of implicit API workflow that you're supposed to initiate a websocket connection with `/ari/events` before calling any endpoint that can refer to the global `websocket`. Bad usage of the API can trigger an unexpected error, but since this is mock/test code we don't care too much. If this was production code, I would complain.

PR #346, opened 2024-09-11
lines API: fix line status after creation

sduthil:
Why:

* line status was left to null until wazo-calld restart


DrPyser (approved):
Could reproduce bug and fix(before/after) successfully.
Well done.

PR #348, opened 2024-09-30
WP 1333 attended transfer moh

DrPyser:
lotta basic refactoring(method extractions, moving stuff around) in the lot.

### DIFF
--- a/integration_tests/suite/test_transfers.py
+++ b/integration_tests/suite/test_transfers.py
@@ -553,14 +553,26 @@ class TestTransfers(RealAsteriskIntegrationTest):
         recipient_channel_id,
         events,
     ):
-        transfer_bridge = self.ari.bridges.get(bridgeId=transfer_id)
-        assert_that(
-            transfer_bridge.json,
-            has_entry(
-                'channels',
-                contains_inanyorder(initiator_channel_id, recipient_channel_id),
-            ),
-        )
+        def receive_transfer_events():
+            assert_that(
+                events.accumulate(with_headers=True),
+                has_items(
+                    has_entries(
+                        message=has_entry('name', 'transfer_abandoned'),
+                        headers=has_entries(
+                            {
+                                'name': 'transfer_abandoned',
+                                'tenant_uuid': VALID_TENANT,
+                            }
+                        ),
+                    ),
+                ),
+            )
+
+        until.assert_(receive_transfer_events, interval=0.5, tries=5)
+
+        self.assert_participants_are_bridged(initiator_channel_id, recipient_channel_id)
+
         assert_that(
             transferred_channel_id,
             self.c.is_hungup(),
@@ -570,55 +582,22 @@ class TestTransfers(RealAsteriskIntegrationTest):
         assert_that(
             initiator_channel_id, self.c.is_talking(), 'initiator channel not talking'
         )
-        assert_that(
-            initiator_channel_id,
-            self.c.has_variable('XIVO_TRANSFER_ID', ''),
-            'variable not unset',
-        )
-        assert_that(
-            initiator_channel_id,
-            self.c.has_variable('XIVO_TRANSFER_ROLE', ''),
-            'variable not unset',
-        )
 
         assert_that(
             recipient_channel_id, self.c.is_talking(), 'recipient channel not talking'
         )
-        assert_that(
-            recipient_channel_id,
-            self.c.has_variable('XIVO_TRANSFER_ID', ''),
-            'variable not unset',
-        )
-        assert_that(
-            recipient_channel_id,
-            self.c.has_variable('XIVO_TRANSFER_ROLE', ''),
-            'variable not unset',
-        )
 
         result = self.calld.get_transfer_result(transfer_id, token=VALID_TOKEN)
-        assert_that(result.status_code, equal_to(404))
+
+        assert_that(result.status_code, equal_to(200))
 
         assert_that(
-            events.accumulate(with_headers=True),
-            has_items(
-                has_entries(
-                    message=has_entry('name', 'transfer_abandoned'),
-                    headers=has_entries(
-                        {
-                            'name': 'transfer_abandoned',
-                            'tenant_uuid': VALID_TENANT,
-                        }
-                    ),
-                ),
-                has_entries(
-                    message=has_entry('name', 'transfer_ended'),
-                    headers=has_entries(
-                        {
-                            'name': 'transfer_ended',
-                            'tenant_uuid': VALID_TENANT,
-                        }
-                    ),
-                ),
+            result.json(),
+            has_entries(
+                {
+                    'id': transfer_id,
+                    'status': 'abandoned',
+                }
             ),
         )
 
@@ -1844,6 +1823,7 @@ class TestTransferFromStasis(TestTransfers):
             self.events,
             tries=5,
         )
+        self.assert_participants_are_bridged(initiator_channel_id, recipient_channel_id)
 
     def test_given_state_ringback_when_transferred_hangup_and_recipient_hangup_then_state_hungup(
         self,
@@ -1990,6 +1970,8 @@ class TestTransferFromStasis(TestTransfers):
             self.events,
             tries=5,
         )
+
+        self.assert_participants_are_bridged(initiator_channel_id, recipient_channel_id)
 
     def test_given_state_abandoned_when_initiator_hangup_then_everybody_hungup(self):
         (
--- a/integration_tests/suite/test_transfers.py
+++ b/integration_tests/suite/test_transfers.py
@@ -167,6 +167,17 @@ class TestTransfers(RealAsteriskIntegrationTest):
             transfer_id,
         )
 
+    def assert_participants_are_bridged(self, *channel_ids):
+        # check for a bridge with both channels
+        all_bridges = self.ari.bridges.list()
+        relevant_bridges = [
+            bridge
+            for bridge in all_bridges
+            if set(channel_ids) <= set(bridge.json['channels'])
+        ]
+        assert relevant_bridges, f"No bridge found for channels {channel_ids}"
+        assert len(relevant_bridges) == 1, "More than one bridge found"
+
     def given_answered_transfer(self, variables=None, initiator_uuid=None):
         (
             transferred_channel_id,
@@ -182,13 +193,6 @@ class TestTransfers(RealAsteriskIntegrationTest):
         transfer_id = response['id']
         recipient_channel_id = response['recipient_call']
 
-        def channel_is_in_bridge(channel_id, bridge_id):
-            return (
-                channel_id in self.ari.bridges.get(bridgeId=bridge_id).json['channels']
-            )
-
-        until.true(channel_is_in_bridge, recipient_channel_id, transfer_id, tries=5)
-
         def transfer_is_answered(transfer_id):
             return self.calld.get_transfer(transfer_id)['status'] == 'answered'
 
@@ -198,6 +202,8 @@ class TestTransfers(RealAsteriskIntegrationTest):
             timeout=5,
             message='transfer was not answered',
         )
+
+        self.assert_participants_are_bridged(initiator_channel_id, recipient_channel_id)
 
         return (
             transferred_channel_id,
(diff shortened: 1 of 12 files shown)


wazo_calld/plugins/transfers/stasis.py
```diff
@@ -140,13 +148,17 @@ def process_answered_calls(self):
...
                     event = {'args': ['', '', transfer.id]}
                     self.transfer_recipient_answered((channel, event))
-        logger.debug('Done.')
+                    answered_calls += 1
+        logger.debug(
+            'Finished processing %d answered calls since last stop.', answered_calls
+        )
 
     def stasis_start(self, event_objects, event):
+        logger.debug('received StasisStart event: %s', event)
```
sduthil, line 161: This is very noisy, since it also logs events unrelated to transfers. Please move the debug messages after verifying the event is related to transfers.

wazo_calld/plugins/transfers/stasis.py
```diff
@@ -170,6 +182,7 @@ def stasis_start(self, event_objects, event):
         self.stasis_start_pubsub.publish(transfer_action, (channel, event))
 
     def hangup(self, channel, event):
+        logger.debug('received StasisEnd/hangup event: %s', event)
```
sduthil, line 185: idem

wazo_calld/plugins/transfers/stasis.py
```diff
@@ -183,34 +196,32 @@ def hangup(self, channel, event):
...
         except KeyError:
             logger.debug('recipient answered, but transfer was abandoned')
-
-            for channel_id in transfer_bridge.json['channels']:
-                try:
-                    ari_helpers.unring_initiator_call(self.ari, channel_id)
-                except ARINotFound:
-                    pass
+            # avoid leaving recipient channel hanging
+            channel.hangup()
```
sduthil, line 211: Why hangup? Shouldn't the initiator and recipient be bridged instead?

wazo_calld/plugins/transfers/stasis.py
```diff
@@ -183,34 +196,32 @@ def hangup(self, channel, event):
         self.hangup_pubsub.publish(transfer_role, transfer)
 
     def transfer_recipient_answered(self, channel_event):
+        logger.debug('received TransferRecipientAnswered event: %s', channel_event)
```
sduthil, line 199: This is message is misleading, since `TransferRecipientAnswered` is not an event

wazo_calld/plugins/transfers/stasis.py
```diff
@@ -183,34 +196,32 @@ def hangup(self, channel, event):
...
-            for channel_id in transfer_bridge.json['channels']:
-                try:
-                    ari_helpers.unring_initiator_call(self.ari, channel_id)
-                except ARINotFound:
-                    pass
+            # avoid leaving recipient channel hanging
+            channel.hangup()
 
     def create_transfer(self, channel_event):
+        logger.debug('received CreateTransfer event: %s', channel_event)
```
sduthil, line 214: This is message is misleading, since `CreateTransfer` is not an event

wazo_calld/plugins/transfers/stasis.py
```diff
@@ -247,6 +258,29 @@ def clean_bridge(self, channel, event):
...
+                )
+                pass
+            else:
+                logger.debug(
+                    'transfer(id=%s) in progress(status=%s) using bridge(id=%s), leaving bridge intact',
+                    transfer_id,
+                    transfer.status,
+                    bridge.id,
+                )
+                return
```
sduthil, line 282: Suggestion for readability, because I find `except / else` to be difficult to read:
```suggestion
            except KeyError:
                logger.debug(
                    'bridge(id=%s) has variable WAZO_TRANSFER_ID=%s, but transfer is not persisted anymore',
                    bridge.id,
                    transfer_id,
                )
                transfer = None
                
            if transfer:
                logger.debug(
                    'transfer(id=%s) in progress(status=%s) using bridge(id=%s), leaving bridge intact',
                    transfer_id,
                    transfer.status,
                    bridge.id,
                )
                return
```

wazo_calld/plugins/transfers/state.py
```diff
@@ -220,59 +308,114 @@ def _start(self, context, exten, variables, timeout):
...
+
+    def _cancel(self):
+        # transfer is cancelled: transfer recipient exits
+        # initiator and transferred channels remain
+        # cancel logic:
+        # 1. unset transfer channel variables
+        # 2. hangup transfer recipient channel
+        # 3. stop ringing on initiator channel
+        # 4. remove transferred party from hold
+        # 5. bridge back transferred and initiator parties
```
sduthil, line 393: The comment about cancel logic seems redundant to me, and risks becoming obsolete at the next change in this function. I suggest removing it, or extracting more logic to functions to make it more readable.

wazo_calld/plugins/transfers/state.py
```diff
@@ -526,40 +593,18 @@ def initiator_hangup(self):
...
-            )['value']
-        except ARINotFound:
-            logger.error('initiator hung up while creating transfer')
-        try:
-            variables = json.loads(variables_str)
-        except ValueError:
-            logger.warning('could not decode transfer variables "%s"', variables_str)
-            variables = {}
-        timeout = None if timeout_str == 'None' else int(timeout_str)
+        # TODO: any error to interpret?
```
sduthil, line 596: I guess `get_initial_transfer_variables` should raise an error if it can't get all variables. Since this case was not handled before either, I don't think it's very important for now.

wazo_calld/plugins/transfers/state.py
```diff
@@ -579,17 +624,19 @@ class TransferStateRingback(TransferState):
     @transition
     def transferred_hangup(self):
         self._abandon()
-        return TransferStateEnded.from_state(self)
+        return TransferStateAbandoned.from_state(self)
```
sduthil, line 627: Oh, nice catch!

wazo_calld/plugins/transfers/state.py
```diff
@@ -663,28 +722,22 @@ def initiator_hangup(self):
 
     @transition
     def recipient_hangup(self):
+        self._abandon()
```
sduthil, line 725: Recipient hanging up is not considered an abandon, but rather a cancel.

wazo_calld/plugins/transfers/state.py
```diff
@@ -790,6 +824,73 @@ def from_state(cls, *args, **kwargs):
         return new_state
 
 
+@state_factory.state
+class TransferStateAbandoned(TransferState):
```
sduthil, line 828: Wow, I'm surprised it did not exist yet... Much better like this :)

wazo_calld/plugin_helpers/ari_.py
```diff
@@ -73,6 +76,31 @@ def get_value():
     raise Exception(f'failed to set channel variable {var}={value}')
 
 
+T = TypeVar('T')
+
+
+class GlobalVariableAdapterProtocol(Protocol[T]):
```
sduthil, line 82: Global variables in Asterisk must be strings, no need for abstract types here.

integration_tests/suite/test_transfers.py
```diff
@@ -561,66 +569,54 @@ def assert_transfer_is_abandoned(
...
-        assert_that(
-            initiator_channel_id,
-            self.c.has_variable('XIVO_TRANSFER_ID', ''),
-            'variable not unset',
-        )
-        assert_that(
-            initiator_channel_id,
-            self.c.has_variable('XIVO_TRANSFER_ROLE', ''),
-            'variable not unset',
-        )
```
sduthil, line 573: Why are those checks removed? They seem still relevant to me

integration_tests/suite/test_transfers.py
```diff
@@ -2293,3 +2276,64 @@ def test_given_answered_transfer_and_transferred_hangs_up_while_calld_is_down_wh
...
+                                'tenant_uuid': VALID_TENANT,
+                                f"user_uuid:{transfer['initiator_uuid']}": True,
+                            }
+                        ),
+                    )
+                ),
+                'transfer_answered event wrong or missing',
+            )
+
+        until.assert_(receive_transfer_events, interval=0.5, tries=5)
```
sduthil, line 2319: Why not use `assert_transfer_is_answered` here?

wazo_calld/plugins/transfers/ari_helpers.py
```diff
@@ -137,3 +140,38 @@ def convert_transfer_to_stasis(
         exten='transfer',
         extra_channel=initiator_call,
     )
+
+
+def get_initial_transfer_variables(
+    ari: ARIClientProxy, initiator_call: str
+) -> tuple[str, str, dict, int | None]:
+    """
+    Dual of convert_transfer_to_stasis above: extract variables injected into initiator_channel
```
sduthil, line 149: What do you mean "dual"?

sduthil (changes requested):

integration_tests/suite/test_transfers.py
```diff
@@ -538,20 +534,32 @@ def assert_transfer_is_blind_transferred(
...
-            transfer_bridge.json,
-            has_entry(
-                'channels',
-                contains_inanyorder(initiator_channel_id, recipient_channel_id),
-            ),
-        )
+        def receive_transfer_events():
+            assert_that(
+                events.accumulate(with_headers=True),
+                has_items(
```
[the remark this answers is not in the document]
DrPyser: Because now, an abandoned transfer does not mean that it has ended yet, as it may be in an intermediary `abandoned` state.
If the expectation is that the transfer is abandoned and the ended, this should require two seperate checks.

integration_tests/suite/test_transfers.py
DrPyser: Hmm, not sure. I feel like I had a reason but can't remember :smh:

wazo_calld/plugins/transfers/ari_helpers.py
DrPyser: https://en.wikipedia.org/wiki/Duality_(mathematics)
"Counterpart", "Complement" are perhaps better suited here.

wazo_calld/plugins/transfers/stasis.py
DrPyser: The transfer doesn't exist anymore, so we can't know the initiator.

And now, as long as the recipient has not answered and the state machine did not process that event, there are no bridges. Only the state machine knows to create the bridge when the participants are ready to talk.
The refactoring also had a goal to encapsulate bridge management within the state machine and decouple other code from implementation details around bridges.

Here we can only assume that if the transfer has ended before the recipient answered, the initiator has/was hung up. If a different happening is possible where the transfer disappears but initiator channel remains, we should handle this explicitly.

wazo_calld/plugin_helpers/ari_.py
DrPyser: The need for generic types is to support the JSON adapter.

wazo_calld/plugins/transfers/stasis.py
sduthil: OK, works for me. In this case, the debug log above is wrong though: the transfer is not "abandoned", it was deleted, right?

wazo_calld/plugin_helpers/ari_.py
sduthil: I see, thanks.

wazo_calld/plugins/transfers/stasis.py
DrPyser: Hmm yes, fair enough it might not be an "abandon" scenario. I can change the log to be more generic.

DrPyser:
> FAILED suite/test_relocates.py::TestCreateUserRelocate::test_given_relocate_ringing_when_initiator_hangs_up_then_all_hangup - AssertionError: relocated not hungup
FAILED suite/test_relocates.py::TestCreateUserRelocate::test_given_relocate_ringing_when_relocated_hangs_up_then_all_hangup - AssertionError: initiator not hungup
FAILED suite/test_relocates.py::TestCreateUserRelocate::test_given_relocate_waiting_completion_when_initiator_hangs_up_then_all_hangup - AssertionError: relocated not hungup
FAILED suite/test_relocates.py::TestCreateUserRelocate::test_given_relocate_waiting_completion_when_relocated_hangs_up_then_all_hangup - AssertionError: initiator not hungup

WHYYY! It's a separate implementation!

DrPyser:
recheck

sduthil (approved):

PR #355, opened 2024-11-12
[WAZO-1908] voicemail: do not move messages to Old directory when recording accessed

bloom1:
Why: it's an undocumented side-effect of a GET, which is not really RESTful since it acts directly on the accessed resource.


DrPyser (approved):
Tested manually, behavior is consistent.

LGTM, Thank you!

PR #369, opened 2025-01-27
transfers: remove hangup locks

sduthil:
Why:

* Since calls are only bridged when strictly necessary, bridges don't
  need to be protected against automatic hangups


sduthil:
recheck

pc-m (approved):

DrPyser:
@sduthil I'm curious, why does this resolve WAZO-4043's issue? Are there additional tests that can be added for regression testing(I assume this would be difficult as this involves a race condition)?

sduthil:
regate

sduthil:
> @sduthil I'm curious, why does this resolve WAZO-4043's issue? Are there additional tests that can be added for regression testing(I assume this would be difficult as this involves a race condition)?

The locking mechanism has special clauses to bypass the lock and terminate one channel that acquired the lock in the first place. In this case, the lock is acquired by the transferred channel, and the transfer is cancelled too early, i.e. before the destination channel is not up/ready/ringing. This causes the lock to be released too early and kill the transferred channel.

fblackburn1 (approved):

PR #398, opened 2026-03-26
WP-2020: Ring group recording

rbienvenu0:
- **recording: Fetch correct channel to stop record**
- **recording: Use channel found instead of the received one**

---

> [!NOTE]
> **Medium Risk**
> Changes call recording channel selection and updates how call direction/bus events treat `Local/` channels, which can affect recording control and call state notifications in production call flows.
> 
> **Overview**
> Fixes recording control for ring group/queue/agent-callback calls by resolving a *real* non-`Local/` ARI channel via `_find_channel_to_record()` and then using `channel.id` (not the incoming `call_id`) for ARI var updates and AMI record start/stop/pause/resume actions.
> 
> Refactors `bus_consume` event relaying to share common channel/call lookup, ensures conversation direction is cached using the resolved channel ID, and filters `Local/` channels out of bridge enter/leave direction updates; adds targeted unit tests for both behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d79b29ee8f81517457a135a992c66dfa0f08e00e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wazo_calld/plugins/calls/bus_consume.py
+++ b/wazo_calld/plugins/calls/bus_consume.py
@@ -276,6 +276,12 @@ class CallsBusEventHandler:
                 return
             participant_channels.append(channel)
 
+        participant_channels = [
+            ch
+            for ch in participant_channels
+            if not ch.json['name'].startswith('Local/')
+        ]
+
         call_direction = self.services.conversation_direction_from_channels(
             self.ari, [channel.id for channel in participant_channels]
         )
@@ -319,6 +325,12 @@ class CallsBusEventHandler:
                 return
 
             participant_channels.append(channel)
+
+        participant_channels = [
+            ch
+            for ch in participant_channels
+            if not ch.json['name'].startswith('Local/')
+        ]
 
         call_direction = self.services.conversation_direction_from_channels(
             self.ari, [channel.id for channel in participant_channels]
--- a/wazo_calld/plugins/calls/bus_consume.py
+++ b/wazo_calld/plugins/calls/bus_consume.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2025 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2016-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
@@ -263,6 +263,8 @@ class CallsBusEventHandler:
             for ch in participant_channels
             if not ch.json['name'].startswith('Local/')
         ]
+        if not participant_channels:
+            return
 
         call_direction = self.services.conversation_direction_from_channels(
             self.ari, [channel.id for channel in participant_channels]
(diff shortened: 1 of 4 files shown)


ja0x (changes requested):
The fix looks sound to me.

Would like to see some code coverage (unit test at least) if possible.

wazo_calld/plugins/calls/services.py
```diff
@@ -732,7 +732,7 @@ def record_start(self, tenant_uuid, call_id):
         if channel_variables['WAZO_CALL_RECORD_ACTIVE'] == '1':
             return
 
-        if self._is_automated_recording(channel_id):
+        if self._is_automated_recording(channel.id):
```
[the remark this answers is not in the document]
rbienvenu0: It's reverted to avoid any side effect, but `_is_automated_recording` is currently a grey area feature-wise so it's hard to tell if we're breaking something or not

wazo_calld/plugins/calls/services.py
```diff
@@ -775,10 +775,7 @@ def record_stop(self, tenant_uuid, call_id):
         if tenant_uuid and channel_helper.tenant_uuid() != tenant_uuid:
             raise NoSuchCall(call_id)
 
-        try:
-            channel = self._ari.channels.get(channelId=channel_id)
-        except ARINotFound:
-            raise NoSuchCall(call_id)
```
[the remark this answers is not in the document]
rbienvenu0: Automated recording is only relevant when starting the call, not when pausing/resuming

wazo_calld/plugins/calls/bus_consume.py
```diff
@@ -276,6 +258,12 @@ def _relay_channel_entered_bridge(self, event):
                 return
             participant_channels.append(channel)
 
+        participant_channels = [
+            ch
+            for ch in participant_channels
+            if not ch.json['name'].startswith('Local/')
+        ]
```
ja0x, line 265:
```suggestion
        ]
        if not participant_channels:
            return None
```
I would propose using a cheap guard after filtering just to make sure we still have usable channels. This protects `conversation_direction_from_channels` to receive an empty list, potentially causing the handler to fail? Haven't looked at the methods code though

wazo_calld/plugins/calls/bus_consume.py
```diff
@@ -320,6 +308,12 @@ def _relay_channel_left_bridge(self, event):
 
             participant_channels.append(channel)
 
+        participant_channels = [
+            ch
+            for ch in participant_channels
+            if not ch.json['name'].startswith('Local/')
+        ]
+
```
ja0x, line 320:
```suggestion
if not participant_channels:
    return None
```
Same here

ja0x (dismissed):
LGTM, manual testing fixes the issue on my end. Added 2 optional small comments about extra protection in bus handlers

wazo_calld/plugins/calls/bus_consume.py
rbienvenu0: Updated

wazo_calld/plugins/calls/bus_consume.py
rbienvenu0: Updated

ja0x (approved):

PR #410, opened 2026-07-03
fix-config-log-filename

fblackburn1:
Why: wazo-calld reads the log file path from the log_filename key;
the logfile key shipped in config.yml was silently ignored, so
overriding it there had no effect.


fblackburn1:
recheck (not related)

PR #423, opened 2026-09-09
[ITEM-644] Group DND persistence

bloom1:
- **ami_helpers: add queue status helper**
- **calls: synchronize users DND status in queue on asterisk boot**

Depends-On: https://github.com/wazo-platform/wazo-auth-keys/pull/105

---

> [!NOTE]
> **Medium Risk**
> Changes telephony queue pause behavior on Asterisk/calld startup and during concurrent DND updates; mistakes could mis-route group calls, though logic is heavily tested and failures are retried.
> 
> **Overview**
> Adds **group DND persistence** so queue member pause state stays aligned with user DND in confd after Asterisk or wazo-calld restarts (pause was previously only in Asterisk memory).
> 
> Introduces **`GroupDNDSynchronizer`**, which compares `QueueStatus` members (`Local/<uuid>@usersharedlines`) with paginated confd DND flags and issues `QueuePause` corrections. Live DND bus events still pause/unpause through the same component, with locking so in-flight sync does not overwrite newer DND changes.
> 
> **`CallsBusEventHandler`** now listens for **`FullyBooted`**, runs sync in a background thread with exponential backoff retries, and triggers sync again on service startup (ARI ready + token refresh). AMI helpers gain **`queue_status`**.
> 
> Integration coverage expands with mock amid controls for queue status/pause failures and delays, plus many scenarios (multi-group members, partial pause, concurrent DND during sync, calld restart).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 178321a5156a3e10c48cccb876f05fa930b88811. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wazo_calld/plugin_helpers/ami.py
+++ b/wazo_calld/plugin_helpers/ami.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2025 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2016-2026 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
@@ -180,6 +180,13 @@ def unpause_queue_member(amid, interface):
     }
     try:
         amid.action('QueuePause', destination)
+    except RequestException as e:
+        raise WazoAmidError(amid, e)
+
+
+def queue_status(amid):
+    try:
+        return amid.action('QueueStatus')
     except RequestException as e:
         raise WazoAmidError(amid, e)
 


wazo_calld/plugins/calls/dnd_synchronizer.py
```diff
@@ -0,0 +1,226 @@
...
+                        user_uuid,
+                        update,
+                    )
+                    applied = update
+            except WazoAmidError as e:
+                logger.warning(
+                    'Failed to correct pause state of user "%s": %s',
+                    user_uuid,
+                    e.details['original_error'],
+                )
```
ja0x, line 164:
```suggestion
            except WazoAmidError as e:
                if e.details['original_error'] != NOT_A_GROUP_MEMBER:
                    raise
                logger.warning(
                    'Member "%s" listed by QueueStatus but unknown to QueuePause, '
                    'skipping',
                    user_uuid,
                )
```
Outages, timeouts and 503 errors all raise this error. Without checking the actual `Interface not found`, a transient amid would prevent any QueuePause from running correctly. Since we never raise or return a exit status, an outage would desync until a full reboot or the system runs long enough to correct itself. Since we never raise, the scan is never retried

wazo_calld/plugins/calls/dnd_synchronizer.py
```diff
@@ -0,0 +1,226 @@
...
+    def _another_run_requested(self):
+        with self._lock:
+            if self._synchronization_requested:
+                self._synchronization_requested = False
+                self._updated_while_synchronizing = {}
+                return True
+            self._synchronizing = False
+            return False
+
+    def _synchronize(self):
```
ja0x, line 112: A member missing from the confd listing defaults to DND-disabled and gets actively unpaused, but the listing is not guaranteed complete. Pages are separate requests ordered only by lastname, and the break at line 194 treats a short listing as authoritative when a user is deleted between pages. On a calld restart, where Asterisk's pause state is already correct, this can unpause a DND user who was fine, silently.

Suggest order='uuid' on both list calls for a total order, and having _fetch_dnd_states report whether it collected total users. When it did not, members absent from the listing should be left alone rather than unpaused. Pausing on partial data is safe, unpausing is not.

(Summarized by Claude)

ja0x (changes requested):

ja0x (approved):

PR #424, opened 2026-09-14
catch call events without tenant_uuid

DrPyser:
> [!NOTE]
> **Medium Risk**
> Changes when call and missed-call events reach the message bus; missing tenant_uuid now drops events that may have been published before, which could affect downstream consumers.
> 
> **Overview**
> Prevents call-related bus events from being published when **`tenant_uuid` is missing**, instead logging a debug line with the call id and **channel name** for diagnosis.
> 
> **`CallNotifier`** now routes most notifications through a shared **`_publish`** helper that no-ops without a tenant; **`user_missed_call`** takes **`tenant_uuid` / `user_uuid` as explicit arguments** (no longer embedded in the payload) and applies the same guard. **`Call`** gains **`channel_name`**, populated when building calls from AMI/ARI (including partial calls for hold/DTMF) so skip logs are easier to trace.
> 
> Adds unit tests for notifier tenant guards and bus handler paths for DTMF/hold/unhold.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8fc436c62ce26f1e34af7c2355dabccf26aa691d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wazo_calld/plugins/calls/notifier.py
+++ b/wazo_calld/plugins/calls/notifier.py
@@ -14,6 +14,7 @@ from wazo_bus.resources.calls.event import (
     MissedCallEvent,
 )
 
+from .call import Call
 from .schemas import call_schema
 
 logger = logging.getLogger(__name__)
@@ -23,40 +24,81 @@ class CallNotifier:
     def __init__(self, bus):
         self._bus = bus
 
-    def call_created(self, call):
+    def call_created(self, channel_name: str, call: Call) -> None:
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_created event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         payload = call_schema.dump(call)
         event = CallCreatedEvent(payload, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_ended(self, call, reason_code):
+    def call_ended(self, channel_name: str, call: Call, reason_code: str) -> None:
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_ended event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         payload = call_schema.dump(call)
         payload.update(reason_code=reason_code)
         event = CallEndedEvent(payload, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_updated(self, call):
+    def call_updated(self, channel_name: str, call: Call) -> None:
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_updated event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         payload = call_schema.dump(call)
         event = CallUpdatedEvent(payload, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_answered(self, call):
+    def call_answered(self, channel_name: str, call: Call) -> None:
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_answered event has no tenant_uuid: %s (%s)',
+                call.id_,
+                channel_name,
+            )
+            return
         payload = call_schema.dump(call)
         event = CallAnsweredEvent(payload, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_hold(self, call):
+    def call_hold(self, channel_name: str, call: Call) -> None:
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_hold event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         event = CallHeldEvent(call.id_, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_resume(self, call):
+    def call_resume(self, channel_name: str, call: Call) -> None:
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_resume event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         event = CallResumedEvent(call.id_, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_dtmf(self, call, digit):
+    def call_dtmf(self, channel_name: str, call: Call, digit: str):
+        if not call.tenant_uuid:
+            logger.debug(
+                'call_dtmf event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         event = CallDTMFEvent(call.id_, digit, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def user_missed_call(self, payload):
-        tenant_uuid = payload.pop('tenant_uuid')
-        event = MissedCallEvent(payload, tenant_uuid, payload['user_uuid'])
+    def user_missed_call(self, payload: dict, tenant_uuid: str, user_uuid: str):
+        if not tenant_uuid:
+            logger.debug(
+                'user_missed_call event has no tenant_uuid: user `%s`', user_uuid
+            )
+            return
+        event = MissedCallEvent(payload, tenant_uuid, user_uuid)
         self._bus.publish(event)
--- a/wazo_calld/plugins/calls/bus_consume.py
+++ b/wazo_calld/plugins/calls/bus_consume.py
@@ -15,6 +15,7 @@ from wazo_calld.plugin_helpers.exceptions import WazoAmidError
 
 from .call import Call
 from .exceptions import NoSuchCall
+from .notifier import CallNotifier
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


wazo_calld/plugins/calls/notifier.py
```diff
@@ -23,40 +24,81 @@ class CallNotifier:
     def __init__(self, bus):
         self._bus = bus
 
-    def call_created(self, call):
+    def call_created(self, channel_name: str, call: Call) -> None:
```
fblackburn1: **[draft]** Threading `channel_name` through 8 public methods purely to print it is worth reconsidering. The information is legitimate — `call.id_` is the `Uniqueid` and does not tell you *what* channel it was — but it belongs on the object already being passed rather than beside it.

`Call` is the domain object here and the name is available everywhere it is built: `channel.json['name']` in `make_call_from_channel`, and via `Channel(channel_id, ari)` in `_partial_call_from_channel_id`. Setting `call.channel_name` there keeps every signature as `(call)`, and the name then shows up on the stasis and services paths too, which this version does not reach.

The cost of the current shape is concrete: it changed 8 signatures and left 3 call sites broken (see the comments below).

`user_missed_call` is the one place where the extra parameters are the right call, for a different reason — `tenant_uuid`/`user_uuid` are routing data the notifier genuinely needs, they were just being smuggled inside the payload.

fblackburn1 (commented):
**[draft]** First-pass review — 5 findings, 2 of them blockers. Details are inline; summary here.

1. **Blocker** — `stasis.py:125` was not updated for the new `call_ended` signature.
2. **Blocker** — `services.py:303` and `:317` were not updated for the new `call_updated` signature.
3. `user_missed_call` no longer pops `tenant_uuid` out of the payload, so it now ships in the event body.
4. The 8 guards are copy-paste, and the predicate is falsy rather than `is None`.
5. Nothing covers the notifier in unit tests, which is why 1 and 2 went unnoticed.

On the log level: this condition raises `ValueError('tenant_uuid must have a value')` from `TenantEvent.__init__` today, which surfaces as `log.exception('Failed to process message')` — an ERROR with a traceback. Dropping to `debug` is a two-level fall to invisible-by-default for a condition that is still abnormal (a call with no tenant means something upstream did not set `WAZO_TENANT_UUID`). `warning` at minimum. The goal is to stop the traceback, not to stop knowing it happened.

On test coverage: `wazo_calld/plugins/calls/tests/test_bus_consume.py` passes `Mock()` as the notifier, so every signature change here is invisible to the unit tests, and there is no `test_notifier.py` at all. A few cases asserting "tenant_uuid absent → nothing published" would cover the new behaviour and catch the two call-site breaks.

wazo_calld/plugins/calls/notifier.py
```diff
@@ -23,40 +24,81 @@ class CallNotifier:
...
+        if not call.tenant_uuid:
```
fblackburn1, line 34: **[draft]** Two smaller things on the guard itself.

The bus rejects on `tenant_uuid is None`; this rejects on any falsy value. An empty string used to be published (`str('')`) and is now silently dropped. That is probably the behaviour you want, but it is a change beyond "catch missing tenant_uuid" and worth being deliberate about.

Also, this block is repeated verbatim 8 times below with only the event name changing. A single private helper — build the event, guard, publish — would leave one place to change the log level or the predicate later.

wazo_calld/plugins/calls/notifier.py
```diff
@@ -23,40 +24,81 @@ class CallNotifier:
...
+            logger.debug(
+                'call_created event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         payload = call_schema.dump(call)
         event = CallCreatedEvent(payload, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_ended(self, call, reason_code):
+    def call_ended(self, channel_name: str, call: Call, reason_code: str) -> None:
```
fblackburn1: **[draft]** Blocker: `stasis.py:125` still calls `self.notifier.call_ended(call, event['cause'])`. With the new signature, `call` binds to `channel_name`, `event['cause']` binds to `call`, and `reason_code` is missing → `TypeError` on every non-local hangup relay. That path has no unit test.

Moving the name onto `Call` (see the comment on `call_created`) would leave this call site untouched.

wazo_calld/plugins/calls/notifier.py
```diff
@@ -23,40 +24,81 @@ class CallNotifier:
...
+                'call_ended event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         payload = call_schema.dump(call)
         payload.update(reason_code=reason_code)
         event = CallEndedEvent(payload, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def call_updated(self, call):
+    def call_updated(self, channel_name: str, call: Call) -> None:
```
fblackburn1: **[draft]** Blocker: `services.py:303` and `services.py:317` still call `self._notifier.call_updated(call)`, so `mute()` and `unmute()` raise `TypeError` — `call` binds to `channel_name` and `call` is left unfilled.

wazo_calld/plugins/calls/notifier.py
```diff
@@ -23,40 +24,81 @@ class CallNotifier:
...
+                'call_dtmf event has no tenant_uuid: %s (%s)', call.id_, channel_name
+            )
+            return
         event = CallDTMFEvent(call.id_, digit, call.tenant_uuid, call.user_uuid)
         self._bus.publish(event)
 
-    def user_missed_call(self, payload):
-        tenant_uuid = payload.pop('tenant_uuid')
-        event = MissedCallEvent(payload, tenant_uuid, payload['user_uuid'])
+    def user_missed_call(self, payload: dict, tenant_uuid: str, user_uuid: str):
```
fblackburn1: **[draft]** Passing `tenant_uuid`/`user_uuid` explicitly is the right move, but dropping `payload.pop('tenant_uuid')` changes the event body: `tenant_uuid` now stays in `data` and reaches websocket consumers as a new field. `tenant_uuid` is already a header, so this looks unintended. `test_calls_bus_consume.py::test_missed_call_event` uses `has_entries`, so it will not catch it.

Either keep the `pop` on the caller side, or stop putting `tenant_uuid` into the dict in `_relay_user_missed_call` in the first place.

wazo_calld/plugins/calls/services.py
```diff
@@ -520,6 +521,7 @@ def channel_destroyed_event(ari, event):
         connected = channel.get('connected')
         caller = channel.get('caller')
         call = Call(channel_id)
+        call.channel_name = channel.get('name')
```
DrPyser, line 524: Why use "get" here and `.json['name']` elsewhere?

wazo_calld/plugins/calls/notifier.py
```diff
@@ -27,60 +30,59 @@ class CallNotifier:
...
+    def user_missed_call(self, payload: dict) -> None:
         tenant_uuid = payload.pop('tenant_uuid')
+        if not tenant_uuid:
+            logger.debug(
+                'user_missed_call event has no tenant_uuid: user `%s`',
+                payload['user_uuid'],
+            )
+            return
         event = MissedCallEvent(payload, tenant_uuid, payload['user_uuid'])
         self._bus.publish(event)
```
DrPyser: why not follow the pattern of other events here?

wazo_calld/plugins/calls/notifier.py
DrPyser: @claude let's keep "is None" check for now.

wazo_calld/plugins/calls/notifier.py
```diff
@@ -27,60 +30,59 @@ class CallNotifier:
...
+        self._publish(CallResumedEvent, call, call.id_)
 
-    def call_dtmf(self, call, digit):
-        event = CallDTMFEvent(call.id_, digit, call.tenant_uuid, call.user_uuid)
-        self._bus.publish(event)
+    def call_dtmf(self, call: Call, digit: str) -> None:
+        self._publish(CallDTMFEvent, call, call.id_, digit)
 
-    def user_missed_call(self, payload):
+    def user_missed_call(self, payload: dict) -> None:
```
DrPyser: @claude let's check why this event has a different interface. how is the payload constructed and can it be built from call and other parameters like other events?

wazo_calld/plugins/calls/notifier.py
```diff
@@ -27,60 +30,59 @@ class CallNotifier:
...
 
-    def call_created(self, call):
-        payload = call_schema.dump(call)
-        event = CallCreatedEvent(payload, call.tenant_uuid, call.user_uuid)
-        self._bus.publish(event)
-
-    def call_ended(self, call, reason_code):
+    def _publish(self, event_class: type[UserEvent], call: Call, *content: Any) -> None:
+        if call.tenant_uuid is None:
+            logger.debug(
```
DrPyser: change this to warning

